PR #42, opened 2026-05-04
Add Makefile and local dev marketplace

rhuss:
## Summary

- Add a `Makefile` with `validate`, `install`, `uninstall`, `test`, `test-e2e`, and `test-all` targets
- Add `.claude-plugin/marketplace.json` to enable persistent local plugin installation

## Why a marketplace.json?

The README documents two installation methods:

1. **Registry:** `claude plugin install agent-eval-harness@opendatahub-skills`
2. **Session-only:** `claude --plugin-dir ./agent-eval-harness`

For local development, the `--plugin-dir` flag only loads the plugin for a single session. To install the plugin *persistently* (so it's available across sessions without extra flags), `claude plugin install` requires a marketplace registry. There is no way to persistently install a local plugin without one.

The `marketplace.json` wraps the existing `plugin.json` (which is unchanged) as a local dev marketplace, so `make install` can register it and install the plugin in one step.

## Usage

```bash
make install     # Register local marketplace + install plugin
make uninstall   # Remove plugin + marketplace
make validate    # Validate plugin manifest
make test        # Unit tests
make test-e2e    # E2E tests (requires ANTHROPIC_API_KEY)
make test-all    # All tests
```

## Test plan

- [x] `make install` registers marketplace and installs plugin
- [x] `make install` is idempotent (clean re-install on repeated runs)
- [x] `make uninstall` removes both plugin and marketplace
- [x] `make validate` passes
- [ ] Verify existing `claude plugin install agent-eval-harness@opendatahub-skills` still works

## Summary by CodeRabbit

* **Chores**
  * Added local development marketplace support to enable installing and managing the agent-eval-harness plugin.
  * Introduced reusable make targets for validation, install/uninstall, and various test runs (unit/e2e/help).

* **New Features**
  * Improved skill discovery: now prioritizes declared skills, and falls back to discovering nested or marketplace-provided skill directories.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,0 +1,22 @@
+{
+  "name": "agent-eval-harness-dev",
+  "owner": {
+    "name": "opendatahub-io"
+  },
+  "metadata": {
+    "description": "Local development marketplace for the agent-eval-harness plugin."
+  },
+  "plugins": [
+    {
+      "name": "agent-eval-harness",
+      "version": "0.1.0",
+      "description": "Agent and skill evaluation harness with MLflow integration",
+      "author": {
+        "name": "opendatahub-io"
+      },
+      "source": "./",
+      "homepage": "https://github.com/opendatahub-io/agent-eval-harness",
+      "repository": "https://github.com/opendatahub-io/agent-eval-harness"
+    }
+  ]
+}
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,37 @@
+.PHONY: validate install uninstall test test-e2e test-all help
+
+MARKETPLACE := agent-eval-harness-dev
+PLUGIN := agent-eval-harness@$(MARKETPLACE)
+
+validate:
+	claude plugin validate ./
+
+install:
+	@claude plugin rm $(PLUGIN) 2>/dev/null || true
+	@claude plugin marketplace rm $(MARKETPLACE) 2>/dev/null || true
+	@claude plugin marketplace add ./ && echo "Marketplace added."
+	claude plugin install $(PLUGIN)
+
+uninstall:
+	@echo "Removing plugin..."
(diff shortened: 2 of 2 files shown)


.claude-plugin/marketplace.json
```diff
@@ -0,0 +1,22 @@
+{
```
astefanutti, line 1: I have been using a symlink from `.claude/skills`. Is a local marketplace a better option from your experience?

astefanutti (commented):
@rhuss thanks! I just have a quick question on how best resolve the skills for local dev otherwise LGTM.

.claude-plugin/marketplace.json
rhuss: For me it was better as I could manage the marketplace/plugins from claude itself (and I can enable/disable the plugin only for certain directories I'm in, keeping a clean context most of the time). My favourite tool for that is --> https://github.com/cc-deck/cc-setup that allows to manage per-user plugins but also per-project. But that only works nicely only when using claude means. Also, sometimes I have also plugins and other stuff for a plugin. Managing symlinks manually becomes tedious quite quickly). 

Btw, I got that idea from rhoai-security-scanner (which behind VPN), and I use this technique for 10+ plugins without issue. I like especally the convenience of a `make install` for publishing an unreleased plugin again.

rhuss: Also, it's easier to find scripts in the skill directory via system reminders (as `CLAUDE_PLUGIN_ROOT` sometimes really behaves weird, see https://github.com/anthropics/claude-code/issues/9354#issuecomment-3584872779)

astefanutti: ok that makes sense, thanks for clarifying.

astefanutti (approved):
LGTM, thanks @rhuss!

PR #43, opened 2026-05-04
Discover skills in nested plugins via marketplace.json

rhuss:
## Summary

- Teaches `find_skills.py` to follow `marketplace.json` source paths to discover skill directories in nested plugins
- Fixes false "skill not found" validator warnings for wrapper projects like [cc-prose](https://github.com/rhuss/cc-prose) where the plugin lives in a subdirectory (e.g., `./prose`)
- Discovery priority: root `plugin.json` skills field, then marketplace source paths, then default dirs

## Test plan

- [ ] Verified against [cc-prose](https://github.com/rhuss/cc-prose): validator reports `VALID` instead of `INCOMPLETE`, all five prose skills discovered
- [ ] Existing behavior unchanged for projects with root-level plugin.json or default skill dirs

Assisted-By: 🤖 Claude Code

## Summary by CodeRabbit

* **Refactor**
  * Improved skill-directory discovery: supports marketplace-discovered nested plugins and flexible config formats (string or list).
  * Safer handling: prevents outside-directory traversal, resolves paths, and filters out non-existent directories.
  * Behavior: respects discovery priority and emits warnings when marketplace reads fail.
  * No user-facing interfaces or commands changed.

### DIFF
--- a/skills/eval-analyze/scripts/find_skills.py
+++ b/skills/eval-analyze/scripts/find_skills.py
@@ -24,6 +24,16 @@ HARNESS_SKILLS = {"eval-setup", "eval-analyze", "eval-dataset", "eval-run",
                    "eval-review", "eval-mlflow", "eval-optimize"}
 
 
+def _resolve_under_cwd(raw, base):
+    """Resolve a path relative to base, rejecting traversal outside CWD."""
+    candidate = (base / Path(raw)).resolve()
+    try:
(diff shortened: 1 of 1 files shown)


astefanutti (approved):
Thanks @rhuss, LGTM!

That makes me think we could also in the future add marketplaces support in eval.yaml to resolve external plugins / skills so eval-analyze can also analyze them.

PR #45, opened 2026-05-04
Add OpenAI-compatible EvalRunner for vLLM/KServe endpoints

Sridhar1030:
## Summary
- Adds an `openai-compatible` runner that evaluates models served via standard OpenAI-compatible `/v1/chat/completions` APIs
- Enables the harness to evaluate models deployed on vLLM, KServe, Ollama, LiteLLM, or any OpenAI-compatible endpoint
- No new dependencies -- uses stdlib `urllib.request`
- Registers in `RUNNERS` dict so `runner.type: openai-compatible` works in eval.yaml

## Motivation
The harness currently only supports Claude Code as a runner. This adds support for evaluating any model behind an OpenAI-compatible API, making the harness useful for a much wider range of deployment targets (vLLM, KServe, Ollama, LiteLLM, etc.).

## Usage
```yaml
# eval.yaml
runner:
  type: openai-compatible
```

Set `OPENAI_BASE_URL` and optionally `OPENAI_API_KEY` in the environment, or pass `base_url`/`api_key` directly.

## Test plan
- 58 unit tests covering constructor validation, request format, success/error paths, ABC compliance, harness integration, and edge cases (null content, empty choices, unicode, large responses)
- Tests use a real local HTTP mock server (not mocked at function level)
- All 111 tests in the repo pass (58 new + 53 existing)

## Release Notes

* **New Features**
  * Introduced support for OpenAI-compatible evaluation runners, enabling evaluation against custom API endpoints with configurable models and authentication.
  * Added token usage tracking, structured error handling, and comprehensive artifact logging of evaluation results.

### DIFF
--- a/agent_eval/agent/openai_compatible.py
+++ b/agent_eval/agent/openai_compatible.py
@@ -28,6 +28,7 @@ class OpenAICompatibleRunner(EvalRunner):
         system_prompt: Optional[str] = None,
         max_tokens: int = 512,
         temperature: float = 0.3,
+        **kwargs,
     ):
         self._base_url = base_url or os.environ.get("OPENAI_BASE_URL", "")
         if not self._base_url:
@@ -109,8 +110,9 @@ class OpenAICompatibleRunner(EvalRunner):
                 exit_code = 1
                 stderr_text = "API returned empty choices array"
             else:
-                response_text = choices[0]["message"]["content"]
-                stdout_lines.append(response_text)
+                response_text = choices[0].get("message", {}).get("content") or ""
+                if response_text:
+                    stdout_lines.append(response_text)
 
             usage = body.get("usage", {})
             if usage:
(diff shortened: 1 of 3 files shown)


agent_eval/agent/openai_compatible.py
```diff
@@ -0,0 +1,189 @@
...
+from typing import Optional
+from urllib.parse import urlparse
+
+from .base import EvalRunner, RunResult
+
+
+_MAX_RESPONSE_BYTES = 50 * 1024 * 1024  # 50 MB hard cap
+
+
+class OpenAICompatibleRunner(EvalRunner):
```
astefanutti, line 24: I wonder whether this could become a Responses API compatible runner so it can execute the skill, like:
https://openai.com/index/equip-responses-api-computer-environment/#agent-skills

I've created #49 which outlines what that could look like.

Sridhar1030: Looked at the PR, 
the Responses API runner with Skills + Shell would be the proper equivalent to ClaudeCodeRunner for agentic skill execution. 
This PR covers the simpler completions use case (scoring raw model output from vLLM/KServe without needing an agentic loop). Happy to keep this as-is for the completions use case and build on top of #49 once it lands.

astefanutti:
@Sridhar1030 I think this has been superseded by #65 and can be closed right?

Sridhar1030:
Yes , closing it

PR #46, opened 2026-05-04
Use isolated venv for Python dependencies

rhuss:
## Summary

- `ensure_deps.py` now creates `.eval-venv/` at the plugin root (using `uv` or stdlib `venv`) instead of running `pip install` into the system python
- New `agent_eval/_bootstrap.py` auto-activates the venv on import, before any third-party deps load
- Scripts that import `agent_eval` get the bootstrap for free via `__init__.py`; yaml-only scripts get an explicit import
- No changes to SKILL.md invocation patterns (`python3 ${CLAUDE_SKILL_DIR}/scripts/...` stays the same)

## Problem

On macOS with Homebrew Python 3.14 (externally-managed environment), `pip install` silently fails. The SessionStart hook reported success, but scripts crashed at runtime:

```
ModuleNotFoundError: No module named 'yaml'
ModuleNotFoundError: No module named 'anthropic'
```

## How the bootstrap works

- **Script invocations** (`python3 script.py`): `_bootstrap` re-execs the process with the venv python via `os.execv()`
- **Inline code** (`python3 -c "from agent_eval..."`): `_bootstrap` patches `sys.path` with the venv's site-packages (can't re-exec since `-c` code isn't in `sys.argv`)
- **Already in venv**: no-op (detected by checking if venv site-packages are on `sys.path`)

## Test plan

- [x] `ensure_deps.py` creates `.eval-venv/` with correct python version
- [x] Second run skips install (stamp caching works)
- [x] `score.py` auto-bootstraps when invoked with system python
- [x] `python3 -c "from agent_eval... import anthropic"` works via sys.path patching
- [x] `find_skills.py` (yaml-only script) bootstraps correctly
- [x] `validate_eval.py` works from cc-prose project directory

Assisted-By: 🤖 Claude Code

## Summary by CodeRabbit

* **New Features**
  * Automatic activation of an isolated evaluation virtual environment for CLI tools and runtime imports; dependency installation now targets that venv and uses a stamp to avoid unnecessary reinstalls.

* **Documentation**
  * Updated setup and health-check guidance to use the isolated venv and clarified manual install fallback steps.

* **Chores**
  * Added the evaluation environment directory to version control exclusions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ __pycache__/
 dist/
 build/
 .venv/
+.eval-venv/
 venv/
 
 # Eval runs and state
--- a/agent_eval/__init__.py
+++ b/agent_eval/__init__.py
@@ -1,3 +1,4 @@
 """Agent Eval Harness — generic evaluation framework for Claude Code skills."""
+import agent_eval._bootstrap  # noqa: F401 — auto-activate venv before third-party imports
 
 __version__ = "0.1.0"
--- a/agent_eval/_bootstrap.py
+++ b/agent_eval/_bootstrap.py
@@ -1,0 +1,41 @@
+"""Auto-activate the eval harness venv if available.
+
+Imported by __init__.py before any third-party deps. Uses only stdlib.
+
+For script invocations (python3 script.py), replaces the process with
+the venv python via os.execv(). For inline code (python3 -c "..."),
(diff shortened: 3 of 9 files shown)


astefanutti (approved):
Thanks @rhuss!

PR #48, opened 2026-05-04
Add venv bootstrap to all Python scripts

rhuss:
## Summary
- Adds `import agent_eval._bootstrap` to all 11 Python scripts that were missing it
- Follow-up to #46 which introduced the bootstrap mechanism but only added it to 4 of 15 scripts
- `score.py` was the critical miss: LLM judges failed with `No module named 'anthropic'` because it ran under system Python without the venv activated

## Test plan
- [x] Verified all 15 Python scripts now have the bootstrap import
- [x] Confirmed `score.py` successfully runs LLM judges when invoked via system `python3`
- [ ] Run full eval suite to verify no regressions

Assisted-By: 🤖 Claude Code

## Summary by CodeRabbit

* **Chores**
  * Evaluation and setup scripts now automatically activate the configured virtual environment at startup, ensuring consistent runtime configuration and more reliable script execution.
  * The package no longer triggers environment activation on import, centralizing startup behavior to the CLI/run-time scripts.

### DIFF
--- a/agent_eval/__init__.py
+++ b/agent_eval/__init__.py
@@ -1,4 +1,3 @@
 """Agent Eval Harness — generic evaluation framework for Claude Code skills."""
-import agent_eval._bootstrap  # noqa: F401 — auto-activate venv before third-party imports
 
 __version__ = "0.1.0"
--- a/skills/eval-mlflow/scripts/attach_feedback.py
+++ b/skills/eval-mlflow/scripts/attach_feedback.py
@@ -25,6 +25,8 @@ Usage:
         --run-id <id> --config eval.yaml --action pull
 """
 
+import agent_eval._bootstrap  # noqa: F401 — auto-activate venv
+
 import argparse
 import os
 import sys
--- a/skills/eval-mlflow/scripts/from_traces.py
+++ b/skills/eval-mlflow/scripts/from_traces.py
@@ -13,6 +13,8 @@ Usage:
         [--min-duration 5]
 """
 
+import agent_eval._bootstrap  # noqa: F401 — auto-activate venv
+
 import argparse
 import sys
 
--- a/skills/eval-mlflow/scripts/log_results.py
+++ b/skills/eval-mlflow/scripts/log_results.py
@@ -11,6 +11,8 @@ Usage:
         --config eval.yaml
 """
 
+import agent_eval._bootstrap  # noqa: F401 — auto-activate venv
+
 import argparse
 import json
 import os
--- a/skills/eval-mlflow/scripts/sync_dataset.py
+++ b/skills/eval-mlflow/scripts/sync_dataset.py
@@ -27,6 +27,8 @@ Schema mapping format (tmp/schema_mapping.json):
     - "filename:__file__" uses the entire file content
 """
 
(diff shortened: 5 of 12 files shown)


astefanutti (commented):
Asking Claude Code which tells me this:

>  The alternative would be to install agent_eval into .eval-venv/ (via pip install -e .), which would put both the package and third-party deps in one place. Then you'd need neither the symlink nor the bootstrap — scripts would just run under the venv python directly. But that means ensure_deps.py needs to do an editable install of the harness itself, and changes to agent_eval/ source wouldn't be reflected until reinstall (unless -e works, which requires the venv python to be used).
> 
>   Actually, pip install -e . in the venv would work: editable installs create a link back to
>   the source, so code changes are reflected immediately. That would let you:
>   1. Drop all symlinks
>   2. Drop _bootstrap.py entirely
>   3. Have ensure_deps.py do pip install -e . -q into the venv alongside the third-party deps
>   4. Have the SessionStart hook invoke scripts with the venv python directly
> 
>   That's cleaner but a bigger change to PR #46. The current approach (symlinks + bootstrap) works, it's just two mechanisms doing what one could handle.

What do you think?

rhuss:
This is what my Claude says :-) -->

The `pip install -e .` approach is cleaner in theory but has practical issues in this context:

1. **Plugin installs may be read-only.** When installed via `claude plugin install`, the plugin root isn't necessarily a writable git checkout. Editable installs need either a `.git` dir or writable source tree for the `.egg-link`/`.pth` mechanism to work.

2. **SKILL.md invocations all use `python3`.** Every `python3 ${CLAUDE_SKILL_DIR}/scripts/...` call across 7 SKILL.md files would need to switch to the venv python. The bootstrap avoids touching any SKILL.md, the scripts just work with whatever python invokes them.

3. **Graceful degradation.** The bootstrap is a no-op when no venv exists (local dev, CI with deps pre-installed). An editable install approach requires the venv python to always be used, which is a coordination problem.

4. **Symlinks already existed.** The `agent_eval` symlinks were there before this PR for PYTHONPATH-free imports. We're not adding a new mechanism, just leveraging what's already in place.

That said, if we later move to a model where the harness is always invoked via a dedicated entry point (not `python3 script.py`), the editable install approach would make sense as a simplification.

rhuss:
@astefanutti but let me quickly check how I do local dep installation in my plugins.

rhuss:
nope, I don't have really a better approach.

astefanutti:
@rhuss thanks, yes that makes sense, we can iterate later.

Just one small point, would that make sense to remove `import agent_eval._bootstrap` from `agent_eval/__init__.py`, the module can be used standalone like in https://github.com/opendatahub-io/agent-eval-harness/blob/main/TRACING.md and I think the bootstrap is only really needed when invoking it from skills. What do you think?

rhuss:
Good point! I've removed the `import agent_eval._bootstrap` from `__init__.py` so the package can be used standalone (e.g. `pip install -e .`, `claude-trace` CLI) without triggering the `os.execv()` side effects. The individual skill scripts already import `agent_eval._bootstrap` directly, so they still get venv activation when invoked by Claude Code.

astefanutti (approved):
Awesome, thanks @rhuss!

PR #73, opened 2026-05-27
feat(eval-optimize): add optimization controls for robust skill improvement

astefanutti:
## Summary

Makes the `/eval-optimize` strategy **selectable** instead of forcing one approach on every run. Adds a SkillOpt-style control set — inspired by [SkillOpt](https://arxiv.org/abs/2605.23904) (Microsoft, 2026), which treats skill documents as trainable external state — for larger datasets, while keeping a lightweight linear flow as the default for the common small-eval case.

Originally this PR *replaced* the simple optimize loop wholesale. Based on review, it now layers the heavy machinery behind a `--strategy` flag that auto-selects by dataset size — so small evals stay cheap and simple, and only larger ones pay for splitting / minibatches / consolidation.

## `--strategy lite|skillopt|auto` (default `auto`)

- **`lite`** — lightweight linear flow: identify failures → analyze → edit → re-run/verify → check regressions → iterate. The pre-overhaul behavior; best for small datasets and quick fixes.
- **`skillopt`** — the full control set below.
- **`auto`** (default) — `lite` for datasets with **< 20 cases**, `skillopt` for ≥ 20. Below ~20 cases the train/selection/test splits and minibatches are statistical noise, so the lightweight flow is cheaper and more reliable.

**One flow, conditional depth:** `lite` is the spine of `SKILL.md`; the SkillOpt machinery is gated behind clearly-marked **"SkillOpt mode"** callouts, so the two paths can't drift apart. Both strategies share always-on safeguards — evidence-grounded edits, regression checks after each edit, and a rejected-edit buffer that prevents retrying failed approaches.

## SkillOpt-mode controls

**Core**
- Data splitting with a validation gate (default `40:20:40`, deterministic seed; small/tiny datasets handled gracefully)
- Bounded, adaptive edit budget (ceiling default 4; the count is chosen per-iteration from the evidence, not a fixed formula)
- Edit ranking by expected utility (systematic impact, complementarity, generality, actionability)

**Reflection & merge**
- Minibatch reflection (parallel sub-agents, groups of 5–8)
- Separate failure vs. success analysis prompts (different cognitive tasks)
- Hierarchical merge with support counts and conflict tracking

**Longitudinal learning**
- Protected slow-update region (`<!-- SLOW_UPDATE_START/END -->`)
- Meta-skill (optimizer-side learning about effective edit patterns; not shipped with the skill)
- Two-phase cross-iteration consolidation + optimization log

**Operational**
- Cumulative cost tracking with a 5× warning threshold
- Rewrite mode as a fallback when patches plateau
- Test-split generalization check
- Early-exit fast path for case-specific-only failures

## Structure

```
skills/eval-optimize/
├── SKILL.md                            # orchestration — lite spine + SkillOpt callouts
├── prompts/
│   ├── failure-analysis.md
│   ├── success-analysis.md
│   ├── merge-proposals.md
│   ├── consolidation.md
│   └── rewrite-skill.md
├── references/optimization-controls.md # control reference + strategy selection
└── scripts/split_dataset.py            # deterministic splitting
tests/test_split_dataset.py
```

## Integration with current main

Rebased onto `main` (1.10.0). The merge preserves main's recent eval-optimize improvements alongside the new flow:
- Config auto-discovery (`--config` optional)
- `$AGENT_EVAL_RUNS_DIR/<eval-name>/<id>` run-path scheme
- Builtin judge-type awareness (`builtin` / `check` / `llm` / `code`)

## What eval-optimize already had that SkillOpt doesn't

- Human feedback integration (`review.yaml` + MLflow annotations)
- Sub-agent delegation for large transcript analysis
- Explicit regression handling with revert capability
- Pipeline composition with `/eval-run`, `/eval-review`, `/eval-mlflow`

## Test plan

- [ ] `/eval-optimize --strategy lite` on a small eval (e.g. rfe-creator dedup, 4 cases) — confirm no splits/minibatches/gate engage
- [ ] `/eval-optimize --strategy skillopt` — full loop: validation gate, consolidation, test split
- [ ] `/eval-optimize` (auto) picks `lite` for < 20 cases and `skillopt` for ≥ 20
- [ ] `split_dataset.py` produces deterministic, reproducible splits; handles `< 3` and `< 10` cases
- [ ] Validation gate rejects neutral edits (ties); rejected-edit buffer prevents retry loops

Ref: https://arxiv.org/abs/2605.23904

### DIFF
--- a/skills/eval-optimize/SKILL.md
+++ b/skills/eval-optimize/SKILL.md
@@ -1,17 +1,23 @@
 ---
 name: eval-optimize
-description: Automated skill improvement loop. Runs eval, identifies failures, edits SKILL.md with evidence-based fixes, re-runs to verify, checks for regressions. Uses data splitting, validation gates, and bounded edits to prevent overfitting. Use when you want to automatically improve a skill, fix failing judges, improve scores, or iterate until tests pass. Triggers on "optimize the skill", "make it pass", "auto-fix", "improve the scores", "why is it failing". Works best after /eval-run.
+description: Automated skill improvement loop. Runs eval, identifies failures, edits SKILL.md with evidence-based fixes, re-runs to verify, checks for regressions. Selectable strategy — a lightweight linear flow for small datasets and a SkillOpt mode (data splitting, validation gate, bounded edits) that prevents overfitting on larger ones, auto-selected by dataset size. Use when you want to automatically improve a skill, fix failing judges, improve scores, or iterate until tests pass. Triggers on "optimize the skill", "make it pass", "auto-fix", "improve the scores", "why is it failing". Works best after /eval-run.
 user-invocable: true
 allowed-tools: Read, Write, Edit, Bash, Glob, Grep, Agent, Skill, AskUserQuestion
 ---
 
-You are an automated skill improver. You treat the skill document as trainable external state: data splitting prevents overfitting, bounded edit budgets prevent over-editing, a validation gate ensures only real improvements are accepted, and a rejected-edit buffer prevents retrying failed approaches. You iterate until judges pass or you hit the max iteration limit.
-
-Why these controls matter: without them, the optimization loop tends to overfit — an edit that helps one failing case breaks three passing ones, and the next iteration tries the same rejected edit again. Data splitting catches overfitting by testing edits on cases the optimizer didn't train on. The edit budget prevents making 10 changes at once (which makes it impossible to isolate what helped). The rejected-edit buffer ensures the loop makes forward progress instead of cycling.
+You are an automated skill improver. You run evaluations, identify what's failing and why, make targeted evidence-based edits to the skill's SKILL.md, re-run to verify, and check for regressions. You iterate until judges pass or you hit the max iteration limit.
 
 The key difference from `/eval-review`: you act autonomously. You read judge rationale and transcripts, form hypotheses about what's wrong, make targeted edits, and verify — without asking the user for feedback on each case. The user sets the goal ("make this pass") and you work toward it.
 
-For the full explanation of each optimization control, see `${CLAUDE_SKILL_DIR}/references/optimization-controls.md`.
+## Strategies
+
+This skill runs in one of two strategies, selected by `--strategy` (default `auto`):
+
+- **`lite`** — a lightweight linear flow: identify failures → analyze root causes → edit → re-run and verify → check regressions → iterate. Best for small datasets (the common case) and quick fixes.
+- **`skillopt`** — treats the skill document as trainable external state, adding controls that prevent overfitting on larger datasets: data splitting with a validation gate, minibatch reflection, a bounded edit budget, cross-iteration consolidation, and a meta-skill. See `${CLAUDE_SKILL_DIR}/references/optimization-controls.md` for the full explanation of each control.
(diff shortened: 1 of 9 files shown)


rhuss:
Thanks for this, Antonin. The SkillOpt-inspired approach is well-researched and addresses real problems I've seen with the current optimizer: thrashing on the same failed edit, overfitting to one case while breaking others, and making too many changes at once to diagnose what helped.

The architectural fit is clean. Pure skill-level change, no core framework modifications. The prompts are well-structured for sub-agent dispatch and the `split_dataset.py` script with deterministic splitting is solid.

**Two questions before merging:**

**1. Has this been run end-to-end?**

The SKILL.md orchestrates a complex flow: data splitting, minibatch analysis with sub-agents, hierarchical merging, validation gating, rejected-edit buffer checks, consolidation, meta-skill updates, and cost tracking. That's a lot for the LLM to follow reliably, and the only mechanically testable piece is `split_dataset.py`. I'd like to understand whether the full loop has been validated on a real skill, and whether the LLM actually respects the validation gate, edit budget, and rejected-edit buffer in practice.

Also: the SKILL.md instructs `/eval-run --case <train_and_selection_cases>` to filter which cases to run. I don't see a `--case` flag in the current `/eval-run` implementation. Is this an assumed dependency?

**2. Would phased delivery reduce risk?**

The 15 controls fall into two natural tiers:

**Tier 1 (core, high confidence):** Data splitting, rejected-edit buffer, bounded edit budget, edit ranking, minibatch analysis, separate failure/success prompts, hierarchical merge. These address the most common optimizer failure modes and are conceptually straightforward.

**Tier 2 (longitudinal, more experimental):** Slow-update protected regions, meta-skill (optimizer-side learning), two-phase consolidation, optimization log, rewrite fallback. These address cross-iteration intelligence, which is valuable but harder to validate.

Tier 2 adds roughly 40% of the SKILL.md complexity (consolidation.md, rewrite-skill.md, meta-skill state, slow-update markers). Shipping Tier 1 first would let us validate the core loop with real usage before layering on the cross-iteration controls.

Not prescriptive, just a suggestion. If you've already validated the full flow end-to-end, shipping it all at once is fine.

rhuss:
A few small skill-authoring suggestions from running the eval-optimize SKILL.md through a pattern analysis:

**1. Sharpen the description's routing boundary**

The description triggers well for optimization intent but doesn't explicitly exclude adjacent skills. Adding an exclusion clause helps Claude route correctly:

```yaml
description: "...Works best after /eval-run. Not for manual case review (use /eval-review) or running evaluations (use /eval-run)."
```

One line, prevents the most common misroute.

**2. Show the final report structure**

The skill produces a report in Step 8 but never shows what it looks like. The LLM has to infer the format, which leads to inconsistent outputs across runs. A brief template would help:

```markdown
### Final Report Template
- **Iterations**: N completed
- **Judges**: [pass/fail summary per judge]
- **Edits applied**: [list with support counts]
- **Edits rejected**: [list with reasons]
- **Test split**: [generalization result]
- **Cumulative cost**: $X.XX
- **Recommendation**: [next steps]
```

**3. Extract inline state management into a script**

The rejected-edit buffer append (Step 5) is inline Python embedded in the SKILL.md:

```python
python3 -c "
import yaml
with open('tmp/rejected-edits.yaml') as f:
    data = yaml.safe_load(f) or {}
data.setdefault('rejected_edits', []).append({...})
..."
```

This pattern appears in a critical path (validation gate rejection). A `scripts/update_state.py` script handling both rejected-edit recording and optimization-log updates would be more robust and testable. Similar to how `split_dataset.py` handles the deterministic splitting.

These are suggestions, not blockers. The skill's core design (data splitting, validation gate, minibatch reflection) is solid.

PR #77, opened 2026-05-28
feat: add prompt-based mode for direct agent evaluation

Prashanth684:
Extends the eval harness to evaluate **agent capabilities directly** — sending a prompt to the agent with no `/skill` wrapper — alongside the existing skill-based evaluation. The flagship use case is **Agentic Documentation Testing**: can an agent navigate your docs, apply your patterns, respect documented constraints, and call your APIs from the documentation alone?

The same work refactors and clarifies several config surfaces: a canonical `execution` block, a unified `generation` provenance block for test-case creation, a builtin `consulted_docs` judge, and a corrected path-based permission guard.

### Prompt mode (`execution.prompt`)
- `execution.prompt: "{{ inputs.prompt }}"` invokes the agent directly — no skill upload, no `/skill` command. Mutually exclusive with `execution.skill`.
- Runs per-case (or batch) through the same workspace → collection → scoring pipeline as skill mode.
- Runner interface unified: `run_skill()` → `execute(target=...)` across `base.py`, `claude_code.py`, `cli_runner.py`; `target=None` selects prompt mode (skip skill upload). `run_skill()` is kept as a deprecated shim.
- Jinja2 rendering of arguments/prompts, plus an `EvalConfig.eval_name()` fallback (skill → name → path → `eval`) so prompt-mode evals derive a run name without a skill.

### Canonical `execution` block
- `execution.skill` is now the canonical way to name a skill; the top-level `skill:` still works but is deprecated (auto-normalized at load).
- `execution.mode` (`case` | `batch`) — *how many* invocations — is orthogonal to `execution.skill` / `execution.prompt` — *what* to execute.

### `generation` block — test-case provenance
Replaces the earlier `dataset.test_categories` taxonomy with a top-level `generation:` block:
- `generation.strategy`: `skill` (default — agent authors cases from skill analysis), `synthetic` (LLM-generated from seeds), or `from-traces` (extracted from MLflow traces).
- For `synthetic`: `generation.context` (repository knowledge injected into every generation prompt) and `generation.seeds` — each seed has a `category`, a required `count`, and exactly one prompt discriminator (mirroring judges): `builtin` (e.g. `docs/navigation`), `prompt_file`, or inline `prompt`. Each seed's `category` is stamped onto cases as `annotations.category`.
- Builtin generation prompts live in `agent_eval/prompts/docs/` (navigation, anti-pattern, authoring, component-usage, architecture), discovered like builtin judges. `list_prompts.py` lists them; `generate_synthetic.py` runs generation.

### `consulted_docs` builtin judge + `tool_trace`
- New builtin **process** judge (`agent_eval/judges/process/consulted_docs.py`) that checks whether the agent read the expected documentation, extracting Read/Grep calls from the execution events — subagent reads included by default.
- New `tool_trace` template variable exposes the tool-call trace to LLM judge prompts.
- The earlier opt-in `traces.documentation_tracking` flag (and `outputs["read_calls"]`) was **removed** in favor of this always-available judge.

### `llm_rubric` scoring shorthand
- `llm_rubric: "..."` on a judge is a concise LLM-evaluation form, with the same context/model overrides as full LLM judges.

### Permission guard fix
- Shared compiler `agent_eval/tools/permissions.py` used by both the local runner and Harbor. Path-based `permissions.deny` rules compile to recursive gitignore patterns (`eval/` → `Read(eval/**)`) for path-scoped tools (`Read`/`Edit`/`Write`/`Grep`/`Glob`); no-op `Bash(path)` patterns are dropped, and a `Bash` deny is hardened into `Read`/`Edit` coverage. Fixes weak non-recursive local rules and the invalid dict rules previously written into Harbor `settings.json`. (Repo-path deny rules protect the local `workspace_mode: repo` answer key; arbitrary subprocess reads still need OS `sandbox.filesystem.denyRead`.)

### Analyze / docs / examples
- `/eval-analyze --prompt <docs>` analyzes documentation and scaffolds a synthetic-generation eval config (`resolve_prompt.py`, analyze-docs prompt).
- New `examples/openshift-agentic-docs.md` recipe + `examples/README.md`.
- New references: `skills/eval-run/references/execution-modes.md`, `skills/eval-dataset/references/synthetic-generation.md`; skill docs refreshed; `skillsaw lint` clean.

### Scoring & robustness fixes
- eval-run honors deny rules, isolates in-repo failures, and validates arguments strictly.
- Scoring fixes for annotations text, doc coverage, and regression + conversation handling.
- `validate_eval` false-positive fixes.

## Example

```yaml
execution:
  mode: case
  prompt: "{{ inputs.prompt }}"

runner:
  type: claude-code
  workspace_mode: repo        # docs navigation needs the real directory tree

generation:
  strategy: synthetic
  context: |
    <repository knowledge injected into every generation prompt>
  seeds:
    - category: navigation
      builtin: docs/navigation
      count: 10

judges:
  - name: consulted_docs
    builtin: consulted_docs
  - name: navigation_success
    prompt: |
      Expected files: {{ annotations.expected_files }}
      Tool usage trace:
      {{ tool_trace }}
      Agent's response:
      {{ conversation }}
    if: "annotations.get('category') == 'navigation'"
```

## How Has This Been Tested?
- `python3 -m pytest tests/ -q` → 526 passed, 2 skipped.
- New suites: prompt analysis, synthetic generation, `llm_rubric` scoring, documentation read-call extraction, permission compilation.
- E2E pipeline (eval-analyze → eval-dataset) against a fake skill fixture.
- `skillsaw lint` clean.

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.


rhuss:
Thanks for this, Prashanth. This is a substantial PR and I took time to understand how it fits architecturally. It genuinely extends the core evaluation pipeline: the config model, runner interface, execution scripts, and scoring all grow consistently to support the new mode.

**What works well:**

- Clean extension of the runner interface (`run_skill()` -> `execute(target: Optional[str])`) with backward-compat shim
- `skill` made `Optional[str]` instead of removed (existing configs unaffected)
- Proper validation (`documentation_tracking` requires `events: true`, mode validation on ExecutionConfig)
- 51 test methods across 4 files
- Taxonomy templates are a smart approach to test case generation
- `llm_rubric` is a useful simplification over full `prompt` templates for straightforward judge criteria
- The in-repo execution mode (`workspace_mode: repo`) with git snapshot safety is well thought out

**The three modes now form a coherent testing spectrum:**
- `case`/`batch`: Does the skill produce correct output when invoked directly?
- `prompt`: Can the agent handle tasks without any skill wrapper?

This naturally leads to a third question: does the right skill *activate* for natural language user intent? We've filed #87 to explore that direction. Your `workspace_mode: repo` infrastructure and the `extract_read_calls()` pattern in events.py provide the foundation that activation testing would build on.

**A few things to address:**

1. **`evalhub/adapter.py` still calls `runner.run_skill()`.** The deprecated shim handles this at runtime, but the adapter should be updated to call `execute()` in this PR to avoid shipping code that uses the deprecated path from day one.

2. **Interaction with #85 (eval directory layout).** Our directory layout spec currently filters discovery by `skill` field presence. Prompt-mode configs with no `skill` won't be discovered. We've acknowledged this as "extend when #77 lands" in the spec. Just flagging so both PRs are aware of the coordination needed.

3. **`llm_rubric` vs `prompt` vs `prompt_file`.** Three LLM judge formats is a lot of surface area. The relationship could be clearer: is `llm_rubric` syntactic sugar that compiles down to a `prompt` internally, or is it a separate code path? If the former, documenting that equivalence would help.

rhuss:
@Prashanth684 I've triggered the bot reviews checks, would be great to address those (and resolve the conflicts). Overall, the PR is really a good match for the framework, and I would recommend to include it.

agent_eval/config.py
```diff
@@ -110,17 +119,27 @@ class ExecutionConfig:
     parallelism: Optional[int] = None
     env: dict = field(default_factory=dict)
 
+    def __post_init__(self):
+        # Validate mode
+        valid_modes = ["case", "batch", "prompt"]
```
astefanutti: I would not conflate the mode and the execution configuration which I think are orthogonal.

I would suggest something like: 

```
# ── Skill mode: arguments pairs with skill ──
  execution:
    mode: case
    skill: rfe.create
    arguments: '--priority {{ input.priority }} "{{ input.prompt }}"'
  # → sends: /rfe.create --priority High "Add signature verification"

# ── Prompt mode: prompt is self-contained, no arguments ──
  execution:
    mode: case
    prompt: "{{ input.prompt }}"
  # → sends: Add signature verification
```

  Same input.yaml feeds both (prompt, priority); one Jinja2 dialect; mode stays case/batch.

agent_eval/config.py
```diff
@@ -58,6 +58,14 @@ class TracesConfig:
     stderr: bool = True    # Capture stderr.log
     events: bool = True    # Parse JSONL into events.json
     metrics: bool = True   # Capture run_result.json metrics
+    documentation_tracking: bool = False  # Capture Read tool calls (opt-in)
```
astefanutti: I would suggest to drop `documentation_tracking` and make doc-consultation a builtin judge:

`read_calls` is just a projection of `events`, which is already a standard trace. `score.py` already derives `conversation` ([L193](#)) and `tool_calls` ([L220](#)) from events on demand at scoring time — `read_calls` can follow the exact same pattern. That makes `traces.documentation_tracking` a redundant flag plus a redundant on-disk artifact.

So this becomes a **net removal** with a small builtin judge as the consumer.

#### Remove

- `agent_eval/config.py` — the `documentation_tracking` field on `TracesConfig` and its `__post_init__` "requires `events: true`" validation
- `skills/eval-run/scripts/collect.py` — the conditional `read_calls.json` generation in `_generate_events_json`
- `skills/eval-run/scripts/score.py` — the conditional `read_calls` loading in `load_case_record` and the `{{ read_calls }}` LLM-judge renderer
- docs mentions in `README.md` / `CLAUDE.md` / `eval-yaml-template.md`

#### Keep

- `agent_eval/events.py` — `extract_read_calls(events)` stays as a pure helper, now called by the judge instead of being gated behind a flag and written to disk

#### Add — `agent_eval/judges/process/consulted_docs.py`

```python
"""Checks the agent read the documentation files listed in expected_files.

Required fields: events, annotations.expected_files
Failure means: The agent did not read enough of the expected documentation.

Arguments:
    min_coverage (float): fraction of expected_files that must be read (default 0.8)
    match (str): path match strategy — "suffix" (default), "exact", or "basename"
"""

from agent_eval.events import extract_read_calls


def _normalize(path, match):
    path = (path or "").replace("\\", "/").strip()
    if match == "basename":
        return path.rsplit("/", 1)[-1]
    return path


def judge(outputs, **kwargs):
    min_coverage = kwargs.get("min_coverage", 0.8)
    match = kwargs.get("match", "suffix")

    expected = [_normalize(p, match)
                for p in outputs.get("annotations", {}).get("expected_files", [])]
    if not expected:
        return (True, "No expected_files specified — nothing to verify")

    # Derive read_calls from events on demand (same pattern as conversation/tool_calls)
    read_calls = extract_read_calls(outputs.get("events", []))
    read = [_normalize(c.get("file_path"), match) for c in read_calls]

    def _hit(exp):
        if match == "suffix":
            return any(r.endswith(exp) or exp.endswith(r) for r in read if r)
        return exp in read

    hits = [e for e in expected if _hit(e)]
    coverage = len(hits) / len(expected)
    passed = coverage >= min_coverage
    return (passed,
            f"Read {len(hits)}/{len(expected)} expected docs "
            f"(coverage {coverage:.0%}, threshold {min_coverage:.0%})")
```

#### Usage

```yaml
traces:
  events: true          # standard trace; the judge reads from it

judges:
  - name: consulted_docs
    builtin: consulted_docs
    if: "annotations.get('category') == 'navigation'"
    arguments:
      min_coverage: 0.8
      match: suffix
```

#### Net effect

The whole `documentation_tracking` feature dissolves into **events (already captured) + one builtin judge** that projects `read_calls` at scoring time. No new config surface, no new on-disk artifact, nothing to validate. It's auto-discovered by `BuiltinJudgeRegistry` and shows up in `list_builtins.py` as `consulted_docs (process/python)`.

skills/eval-analyze/SKILL.md
```diff
@@ -175,3 +233,195 @@ If validation produced warnings, list them so the user knows what's incomplete.
...
+
+5. **Write eval.yaml**:
+```bash
+# Write the generated config
+Write(config_path, generated_yaml_content)
+```
+
+#### Example: Documentation Analysis
+
+The builtin `docs` prompt (`prompts/analyze-docs.md`) generates:
```
astefanutti: It's already in `analyze-docs.md`, probably better remove the duplication and keep SKILL.md generic.

skills/eval-analyze/prompts/analyze-docs.md
astefanutti, line 1: ### `prompts/analyze-docs.md` (583 lines) — slim it hard, most of it dissolves

This file is the biggest version of the issues raised elsewhere in the PR. Once the builtin-judge and dedup points land, most of it goes away.

#### What's in it

| Steps | Content | Verdict |
|---|---|---|
| 1–3 | Discover docs, classify repo (Type A/B/C), extract `domain` (constraints/APIs/components) | **Keep** — the real recipe intelligence |
| 4 | Suggest test categories from the builtin list | Keep, slim |
| 5 | Generate dataset schema per category | Keep, condense |
| 6 | Generate judges with embedded `llm_rubric` examples | **Cut** → builtin judges |
| 7 | Full `eval.yaml` template | **Cut** → duplicates SKILL.md example + `eval-yaml-template.md` |

#### What dissolves once the other threads land

- **Step 6 mostly evaporates.** The generated `llm_rubric` judges (`consulted_docs`, `navigation-accuracy`, …) are exactly what should be **builtin judges**. The recipe then references `builtin: consulted_docs` instead of generating rubric prose. Largest single cut.
- **Step 7 goes away** via dedup — point to the shared reference instead of inlining ~80 lines of YAML.
- The triple-duplicated judge examples go with Step 6.

#### Deeper smell: schema enforcement in prose

The file is full of `CRITICAL: only use these builtin templates`, `MUST use count not target_cases`, `NEVER use scoring: or validation:`. That's the prompt doing a validator's job — a symptom of a fragile interface. Fix it upstream, not with louder prose:

- Schema errors → `validate_eval.py` / `generate_from_taxonomy.py` should fail with clear messages.
- Template names → a `list_templates.py` mirroring the existing `list_builtins.py`, so the recipe *discovers* valid names instead of hardcoding "only use these."

Every `CRITICAL:` you can delete by making the interface self-describing is a win.

#### De-hardcode

Model names (`claude-sonnet-4-6` / `claude-opus-4-6`) and OpenShift-specific examples (`MachineConfig`, `openshift/enhancements`) are baked in. Models go stale; the examples make a "generic" recipe look domain-locked. Mark them illustrative or pull from config.

#### Net

After cutting Steps 6–7, deduping, and moving schema enforcement out of prose, this realistically drops from **583 → ~150–200 lines**: discover docs → classify repo → extract domain → pick categories → emit a minimal config that *references* builtins and the shared template.

#### Meta-question

Even slimmed, this is **one opinionated domain recipe**. If `--prompt`/`--recipe` is a pluggable seam, `docs` can be the *reference* recipe — but a 583-line, OpenShift-flavored methodology probably shouldn't ship as a blessed `builtin:` in core `eval-analyze`. It reads more like an **example recipe** users adapt than core infrastructure. Suggest: keep the seam, slim the recipe, and reconsider whether `builtin:docs` belongs in core or as a documented example.

astefanutti (commented):
Thanks @Prashanth684, I've left a couple of comments.

rhuss:
@Prashanth684 Wanted to connect Antonin's feedback here with the activation testing direction in [#87](https://github.com/opendatahub-io/agent-eval-harness/issues/87), because they reinforce each other and together suggest a cleaner path forward.

### The common thread: don't fork modes, extend the config shape

Antonin's suggestion to keep `mode: case` and distinguish skill vs prompt execution by which fields are present (comment on `config.py:124`) naturally extends to activation testing:

```yaml
# Skill mode (existing)
execution:
  mode: case
  skill: rfe.create
  arguments: '--priority {{ input.priority }} "{{ input.prompt }}"'

# Prompt mode (this PR)
execution:
  mode: case
  prompt: "{{ input.prompt }}"

# Activation mode (#87, future)
execution:
  mode: case
  prompt: "{{ input.prompt }}"
  workspace_mode: repo
```

One pipeline, different config shapes. Each mode adds a field rather than forking the execution path.

### The builtin judge pattern scales

Antonin's proposal to replace `documentation_tracking` with a `consulted_docs` builtin judge (deriving `read_calls` from events at scoring time) is exactly the same pattern #87 needs:

- **This PR**: `extract_read_calls(events)` -> `consulted_docs` builtin judge
- **Issue #87**: `extract_skill_calls(events)` -> `activation_correct` builtin judge

Both are projections of the event stream, consumed by judges, no new config flags or on-disk artifacts needed. Building the `consulted_docs` judge now establishes the pattern that activation mode reuses directly.

### Practical suggestion

If you adopt Antonin's feedback (no new mode, builtin judge instead of `documentation_tracking`), the PR gets smaller and the remaining delta for #87 becomes just:
1. `extract_skill_calls()` in `events.py` (same shape as `extract_read_calls()`)
2. Two builtin judges (`activation_correct`, `no_collision`)
3. `workspace_mode: repo` support in workspace.py

Happy to help with any of this. Let us know if you're still working on it or if you'd like to pair on the rework.

skills/eval-analyze/SKILL.md
Prashanth684: removed

agent_eval/config.py
Prashanth684: makes sense - i have made the change - although for prompt mode i'm not sure what batch processing would look like or it would even be feasible.

Prashanth684:
> @Prashanth684 Wanted to connect Antonin's feedback here with the activation testing direction in [#87](https://github.com/opendatahub-io/agent-eval-harness/issues/87), because they reinforce each other and together suggest a cleaner path forward.
> 
> ### The common thread: don't fork modes, extend the config shape
> Antonin's suggestion to keep `mode: case` and distinguish skill vs prompt execution by which fields are present (comment on `config.py:124`) naturally extends to activation testing:
> 
> ```yaml
> # Skill mode (existing)
> execution:
>   mode: case
>   skill: rfe.create
>   arguments: '--priority {{ input.priority }} "{{ input.prompt }}"'
> 
> # Prompt mode (this PR)
> execution:
>   mode: case
>   prompt: "{{ input.prompt }}"
> 
> # Activation mode (#87, future)
> execution:
>   mode: case
>   prompt: "{{ input.prompt }}"
>   workspace_mode: repo
> ```
> 
> One pipeline, different config shapes. Each mode adds a field rather than forking the execution path.
> 
> ### The builtin judge pattern scales
> Antonin's proposal to replace `documentation_tracking` with a `consulted_docs` builtin judge (deriving `read_calls` from events at scoring time) is exactly the same pattern #87 needs:
> 
> * **This PR**: `extract_read_calls(events)` -> `consulted_docs` builtin judge
> * **Issue [feat: Add skill activation testing mode (execution.mode: activation) #87](https://github.com/opendatahub-io/agent-eval-harness/issues/87)**: `extract_skill_calls(events)` -> `activation_correct` builtin judge
> 
> Both are projections of the event stream, consumed by judges, no new config flags or on-disk artifacts needed. Building the `consulted_docs` judge now establishes the pattern that activation mode reuses directly.
> 
> ### Practical suggestion
> If you adopt Antonin's feedback (no new mode, builtin judge instead of `documentation_tracking`), the PR gets smaller and the remaining delta for #87 becomes just:
> 
> 1. `extract_skill_calls()` in `events.py` (same shape as `extract_read_calls()`)
> 2. Two builtin judges (`activation_correct`, `no_collision`)
> 3. `workspace_mode: repo` support in workspace.py
> 
> Happy to help with any of this. Let us know if you're still working on it or if you'd like to pair on the rework.

@rhuss @astefanutti thanks so much for the prompt reviews. Agree completely — extending config shape instead of forking modes is cleaner and more composable. I have made the changes to not have  prompt as part of mode to accommodate activation testing. What has been done:

1. What to execute determined by field presence
```
has_skill = bool(self.skill and self.skill.strip())
has_prompt = bool(self.prompt and self.prompt.strip())
```
```
if has_skill and has_prompt:
      raise ValueError(
          "execution.skill and execution.prompt are mutually exclusive. "
          "Use skill for '/skill-name' invocations or prompt for direct prompts."
)
```   
2. Key separation achieved:
  - mode (case/batch) = how many invocations
  - skill vs prompt field presence = what to execute
  
 3. Backward-compatible eval_name() derivation:
  - Priority chain: skill → name → path → "eval"
  - Added config_path field for path-based derivation

I'll look at the builtin judge pattern next.

skills/eval-analyze/prompts/analyze-docs.md
Prashanth684: > ### `prompts/analyze-docs.md` (583 lines) — slim it hard, most of it dissolves
> This file is the biggest version of the issues raised elsewhere in the PR. Once the builtin-judge and dedup points land, most of it goes away.
> 
> #### What's in it
> Steps	Content	Verdict
> 1–3	Discover docs, classify repo (Type A/B/C), extract `domain` (constraints/APIs/components)	**Keep** — the real recipe intelligence
> 4	Suggest test categories from the builtin list	Keep, slim
> 5	Generate dataset schema per category	Keep, condense
> 6	Generate judges with embedded `llm_rubric` examples	**Cut** → builtin judges
> 7	Full `eval.yaml` template	**Cut** → duplicates SKILL.md example + `eval-yaml-template.md`
> #### What dissolves once the other threads land
> * **Step 6 mostly evaporates.** The generated `llm_rubric` judges (`consulted_docs`, `navigation-accuracy`, …) are exactly what should be **builtin judges**. The recipe then references `builtin: consulted_docs` instead of generating rubric prose. Largest single cut.
> * **Step 7 goes away** via dedup — point to the shared reference instead of inlining ~80 lines of YAML.
> * The triple-duplicated judge examples go with Step 6.
> 
> #### Deeper smell: schema enforcement in prose
> The file is full of `CRITICAL: only use these builtin templates`, `MUST use count not target_cases`, `NEVER use scoring: or validation:`. That's the prompt doing a validator's job — a symptom of a fragile interface. Fix it upstream, not with louder prose:
> 
> * Schema errors → `validate_eval.py` / `generate_from_taxonomy.py` should fail with clear messages.
> * Template names → a `list_templates.py` mirroring the existing `list_builtins.py`, so the recipe _discovers_ valid names instead of hardcoding "only use these."
> 
> Every `CRITICAL:` you can delete by making the interface self-describing is a win.
> 
> #### De-hardcode
> Model names (`claude-sonnet-4-6` / `claude-opus-4-6`) and OpenShift-specific examples (`MachineConfig`, `openshift/enhancements`) are baked in. Models go stale; the examples make a "generic" recipe look domain-locked. Mark them illustrative or pull from config.
> 
> #### Net
> After cutting Steps 6–7, deduping, and moving schema enforcement out of prose, this realistically drops from **583 → ~150–200 lines**: discover docs → classify repo → extract domain → pick categories → emit a minimal config that _references_ builtins and the shared template.
> 
> #### Meta-question
> Even slimmed, this is **one opinionated domain recipe**. If `--prompt`/`--recipe` is a pluggable seam, `docs` can be the _reference_ recipe — but a 583-line, OpenShift-flavored methodology probably shouldn't ship as a blessed `builtin:` in core `eval-analyze`. It reads more like an **example recipe** users adapt than core infrastructure. Suggest: keep the seam, slim the recipe, and reconsider whether `builtin:docs` belongs in core or as a documented example.

@astefanutti thanks for this feedback. spent some time on this and i restructured the analyze-docs.md to be more of a generic documentation evaluation guide, with the openshift specific guide in examples/. Was also able to add more validation rather than prompted guidance through "CRITICAL" statements.

Prashanth684:
@astefanutti @rhuss incorporated the feedback. Please have a look and let me know if i have missed anything!

rhuss:
@Prashanth684 I will have a look today, but could check the conflict and test failures in the meantime ?

agent_eval/config.py
```diff
@@ -170,18 +180,39 @@ class ExecutionConfig:
...
+
+        # Validate skill/prompt mutual exclusivity
+        has_skill = bool(self.skill and self.skill.strip())
+        has_prompt = bool(self.prompt and self.prompt.strip())
+
+        if has_skill and has_prompt:
+            raise ValueError(
+                "execution.skill and execution.prompt are mutually exclusive. "
+                "Use skill for '/skill-name' invocations or prompt for direct prompts."
+            )
```
rhuss, line 289: Consider validating that at least one of `skill` or `prompt` is set. Currently a config with neither passes validation here but fails later in execute.py with a less obvious error. Something like:

```python
if not has_skill and not has_prompt:
    raise ValueError(
        "execution requires either 'skill' or 'prompt'. "
        "Use skill for '/skill-name' invocations or prompt for direct prompts."
    )
```

Not blocking, just earlier feedback for users.

agent_eval/judges/process/consulted_docs.py
```diff
@@ -0,0 +1,44 @@
+"""Checks the agent read the documentation files listed in expected_files.
+
+Required fields: events, annotations.expected_files
+Failure means: The agent did not read enough of the expected documentation.
+
+Arguments:
+    min_coverage (float): fraction of expected_files that must be read (default 0.8)
+    match (str): path match strategy — "suffix" (default), "exact", or "basename"
+"""
```
rhuss, line 19: Worth noting in the docstring that this only tracks `Read` tool calls. If an agent uses `Bash` with `cat`/`head`/`tail` to read files, those won't appear in the coverage check. Users should know this when setting `min_coverage` thresholds, since some agents (especially non-Claude Code runners) may prefer shell-based file access.

skills/eval-analyze/prompts/analyze-docs.md
```diff
@@ -0,0 +1,374 @@
+# Documentation Analysis Prompt for eval-analyze
```
rhuss: Good progress slimming this down (583 → 374 lines). Steps 6-7 correctly reference builtins and the template now, and the hardcoded models and OpenShift examples are gone.

Two remaining areas to tighten:

**Steps 4-5 are still repetitive.** Each repo type gets a full YAML block with test_categories, then Step 5 repeats full schema blocks per category. Since `list_templates.py` exists now, the prompt can discover templates at runtime. Consider collapsing to one example block plus "adapt categories to your repo type" guidance.

**The Type A/B/C classification doesn't generalize.** The three types and five templates work for the repos they were designed around, but common shapes don't fit cleanly:

- **Monorepos** that are A+B+C simultaneously (enhancement proposals + component APIs + runbooks). Forcing a single type means missing categories relevant to the other facets.
- **Infrastructure-as-code** (Terraform, Helm, Kustomize) where the "documentation" is the config files themselves and testing is about declarative composition, not navigating markdown.
- **Operational runbook repos** where agents need to follow diagnostic decision trees, not find APIs or apply design patterns. None of the five templates capture "can the agent follow a troubleshooting flowchart."
- **ML/data pipeline repos** where understanding data flow and transformation semantics doesn't map to any of the three types.
- **Config-driven repos** (CI/CD, GitHub Actions) where testing is "can the agent compose a pipeline stage correctly."

This isn't a blocker, but it reinforces Antonin's suggestion: the taxonomy is one opinionated recipe, not a universal framework. Shipping as `builtin:docs` in core implies generality it doesn't have yet. Consider moving to `examples/` as a reference recipe users adapt, keeping the `--prompt` seam generic for other taxonomies.

rhuss (commented):
Thanks for incorporating the feedback, Prashanth. The core architectural changes look solid. Quick summary of what I checked:

**Addressed well:**
- Mode shape: `execution.mode` is strictly `case`|`batch`, with `skill`/`prompt` as separate mutually exclusive fields. Exactly the pattern we discussed.
- `documentation_tracking` fully removed, replaced by `consulted_docs` builtin judge deriving `read_calls` from events at scoring time.
- `evalhub/adapter.py` migrated to `execute()` (no more deprecated `run_skill()` path).
- `llm_rubric` clearly documented as syntactic sugar compiling to the same internal prompt. `score.py` L920-921 makes the relationship explicit.
- `list_templates.py` added for self-describing template discovery.

**Remaining items:**
- `analyze-docs.md` still at 374 lines (see inline comment with specifics).
- Merge criteria checkboxes are unchecked.
- The `AGENTS.md:50` skillsaw lint violation ("correctly") is on pre-existing content, not introduced by this PR, so not a blocker here.

skills/eval-analyze/prompts/analyze-docs.md
```diff
@@ -0,0 +1,374 @@
...
+```
+
+For each documentation area found above, list the structure:
+
+```bash
+# For each discovered area (ai-docs, docs, guidelines, etc.):
+find <doc-area> -name "*.md" | head -30
+```
+
+Categorize directories by purpose:
```
rhuss: this taxonomy is very specific to how ai-docs are structured, another data point to move it to an example or keep it in a separate repo.

skills/eval-analyze/prompts/analyze-docs.md
Prashanth684: @rhuss thanks again for this feedback - i've been thinking about this a little bit an made some changes
- made prompt mode generic without any `builtin:analyze-docs` - like you said it seems too prescriptive and given that the Openshift specific prompt has been moved to `examples/`, i didn't see a need for this anymore.
- rather than have builtin test_categories, moved it to be a documentation folder instead just to make it consistent and indicate that it is specific.

I think this makes the framework more generic for prompt mode. please let me know your thoughts and if it looks good

Prashanth684:
@rhuss @astefanutti i believe all comments have been addressed and the prompt mode is as generic as can be. Please review and let me know if it is good to merge.

astefanutti:
## Follow-up fixes (pushed to this branch)

I ran a review of this PR and pushed 5 commits addressing the findings. All 498 unit tests pass and each fix was verified behaviorally. Commits are scoped so they're easy to review independently.

### Correctness — wrong / silent execution

- **Skill resolution across substrates** (`542bda8`): the PR moved the skill name to `execution.skill` and made top-level `skill` optional, but only the local orchestrator learned the new resolution order. **EvalHub** (`adapter.py`) and **Harbor** (`tasks.py`) still read `config.skill` (→ `None` for `execution.skill`-only configs), so they silently ran an empty prompt with no `/skill` wrapper instead of the skill under test. Added `EvalConfig.resolve_skill()` / `is_prompt_mode()` as one shared resolver and routed all three substrates through it (incl. proper prompt-mode rendering). `eval_name()`, `log_results.py`, and `config_translator.py` now use it too.

- **`deny` rules stripped in most modes** (`60381f6`): `_filter_permissions_for_mode` removed user-authored `permissions.deny` for every non-in-repo run (all batch + isolated case runs) and its in-repo detection failed open. A config with `deny: ["WebFetch"]` silently ran with denies dropped. Now honored in all modes via `_resolve_permissions()`; repo-protection denies simply don't match isolated `/tmp` paths.

- **One dirty case aborted the whole run** (`60381f6`): `_run_single_case_in_repo` raised an uncaught `RuntimeError` on a dirty repo (and `git status` ran with `check=True`), discarding completed cases and the summary. A repo-modified/errored case is now recorded as a failed case and the run continues.

- **Non-strict Jinja2 args** (`60381f6`, `542bda8`): `_resolve_arguments` rendered with the default lenient `Undefined`, so a missing required field rendered to empty silently despite the docstring promising `ValueError`. Now uses `StrictUndefined`.

- **`{{ annotations }}` regressed to a dict repr** (`906c751`): changed from formatted text to a raw dict, silently degrading existing judge prompts (docs still described the text form). Wrapped in `_AnnotationsProxy` so `{{ annotations }}` renders text again **and** `{{ annotations.get('category') }}` works.

- **`consulted_docs` coverage inflation** (`906c751`): an empty `""` entry in `expected_files` matched every read, and bidirectional `endswith` produced substring false positives (`report.md` ~ `final-report.md`). Now drops empty entries and matches only at path-component boundaries.

- **Regression thresholds silently skipped** (`906c751`): `detect_regressions` skipped `min_pass_rate`/`min_win_rate`/`min_mean` when the metric was `None` (numeric judge, or all cases skipped). Now surfaced so the unevaluated threshold isn't hidden.

### Correctness — validator rejecting valid configs (`e04688b`)

`validate_eval.py` rejected valid, runnable configs with `sys.exit(1)`, blocking `/eval-analyze`:
- `standard_vars` omitted `arguments` / `outputs` / `annotations_text` (all provided at scoring time), and the extraction regex flagged loop variables, filters, and Jinja2 globals (`range`) as undefined → now extracted via `jinja2.meta.find_undeclared_variables`.
- test-render `mock_data` lacked `outputs`/`arguments`/`annotations_text` and the `tojson` filter → added.
- `valid_judge_fields` omitted `samples` and `feedback_type` → added.
- `entry_point` existence was checked relative to CWD instead of the eval.yaml dir → resolved via `config_dir`.
- `builtin:` template refs validated but crashed at generation after the `builtin/` → `documentation/` rename → `resolve_template` now maps the deprecated prefix.

### Lower severity

- `run_skill` deprecated shim now forwards `extra_env` (`3b3c16a`).
- Conversation auto-wrap matched an exact-spacing literal, double-wrapping `{{conversation}}` / `{{  conversation  }}` → now matches any spacing (`906c751`).
- **Pre-existing** (from `main`, but this PR adds a new call site): `inject_tracing_env` / `inject_tracing_hook` wrote MLflow vars under the settings `"environment"` key, which Claude Code ignores — it reads `"env"` (as `workspace.py` / `tools.interception` do). Fixed to `"env"` (`3b3c16a`).

### Note on one finding I did *not* change

`_load_llm_judge` prefers `llm_rubric` when both `prompt` and `llm_rubric` are set, while the validator errors on "multiple implementations." I kept the validator strict on purpose — rejecting an ambiguous config up front is the safer contract, and since validation gates runtime the disagreement can't cause a wrong run.

astefanutti:
## Follow-up: `execution.skill` is now canonical, top-level `skill` deprecated (`15a937f`)

Per discussion, completed the migration this PR started rather than leaving the dual-home in place. `execution.skill` is now the single canonical location (symmetric with `execution.prompt` / `execution.mode`), and the top-level `skill:` field is deprecated **but still fully supported for external projects**:

- `from_yaml` continues to auto-normalize top-level `skill:` → `execution.skill`, and now emits a `DeprecationWarning`.
- `validate_eval` emits a matching migration WARNING when top-level `skill:` is used without `execution.skill`.
- Skill-name path-segment validation and discovery eval-name derivation now use the *resolved* skill, so `execution.skill`-only configs behave identically to the old top-level form.
- Remaining display/log spots (evalhub adapter/runner, harbor CLI, `check_env`) go through `resolve_skill()`.
- Migrated in-repo config + docs (`eval.yaml`, `README.md`, `CLAUDE.md`) to `execution.skill`; the `eval-yaml-template.md` the analyze skill generates from already used it.
- Added tests for the deprecation warning, auto-normalization, and prompt-mode resolution. Full suite: **501 passed**.

External `eval.yaml` files using top-level `skill:` keep working unchanged (with a warning) until removal in a future major version. The switch itself lives entirely inside `EvalConfig.resolve_skill()` / `is_prompt_mode()`, so removing the fallback later is a one-line change.

Prashanth684:
hey @astefanutti thanks for the fixes - those were very helpful. i have added some more cminor cleanups and enhancements:

 **New feature: {{ tool_trace }} template variable for LLM judges**

  - agent_eval/events.py — adds extract_tool_trace() function that renders a chronological, human-readable log of all tool calls (Read, Bash, Agent, Edit, Grep, etc.) with formatted inputs; also adds _format_tool_input() helper with per-tool formatting logic
  - skills/eval-run/scripts/score.py — wires {{ tool_trace }} into Jinja2 rendering so LLM judges can see agent behavior (navigation, exploration), not just text output

  **Enhanced consulted_docs builtin judge**

  - agent_eval/events.py — extract_read_calls() now accepts include_grep=True to count Grep tool calls as file reads
  - agent_eval/judges/process/consulted_docs.py — adds include_grep and preloaded_files arguments; preloaded files (e.g. CLAUDE.md) count as "already consulted" without needing an explicit Read call; detail string notes auto-loaded hits
  - skills/eval-analyze/scripts/validate_eval.py — registers include_grep and preloaded_files as valid builtin arguments for consulted_docs

  **Documentation templates: fix input/annotations split**

  - 5 template files (navigation.md, anti-pattern.md, architecture.md, authoring.md, component-usage.md) — move annotation fields (expected_files, expected_components, category, etc.) out of input.yaml into annotations.yaml where they belong; rename expected_documentation → expected_files for consistency

  **Expanded taxonomy annotation fields**

  - skills/eval-dataset/scripts/generate_from_taxonomy.py — adds new recognized annotation fields (expected_constraint, expected_structure, expected_patterns, expected_api, expected_example_type, expected_fields, expected_components, expected_interactions, correct_approach) to match what the templates actually
  produce

  **Minor cleanups**

  - skills/eval-run/scripts/execute.py / workspace.py — use config.is_prompt_mode() and config.resolve_skill() instead of inlining the logic
  - skills/eval-run/scripts/score.py — load_case_record() now reads per-case metrics from run_result.json's per_case dict with fallback to top-level
  - README.md, agent_eval/config.py, docs/references — document {{ tool_trace }} everywhere prompts are described
  - examples/openshift-agentic-docs.md — updated example to show include_grep, preloaded_files, and {{ tool_trace }} usage

astefanutti:
I've rebased this branch onto latest `main` and did some design changes:

**1. `dataset.test_categories` → a top-level `generation:` block.** Splits "what the dataset is" from "how cases are generated":
- `generation.strategy`: **`skill`** (default) · **`synthetic`** · **`from-traces`** — one provenance axis. Absent = `skill`, so existing skill configs are untouched.
- `generation.seeds` (was `test_categories`) and `generation.context` (was `dataset.domain`). Each seed has a `category`, a required `count`, and exactly one prompt discriminator: `builtin` / `prompt_file` / `prompt` — mirroring how judges work.

**2. Builtin generation prompts now mirror builtin judges.** Moved `skills/eval-dataset/templates/documentation/*` → `agent_eval/prompts/docs/*`, added a `BuiltinPromptRegistry` + `resolve_seed_prompt()`, and a `list_prompts.py` discovery script (parallel to `list_builtins.py`). Renamed `generate_from_taxonomy.py` → `generate_synthetic.py`; dropped `resolve_template.py`/`list_templates.py`.

**3. Unified provenance; dropped the CLI `--strategy` flag.** `from-traces` moved from a CLI value into `generation.strategy` (beside `synthetic`), and bootstrap-vs-expand is now derived from the current dataset state rather than a flag (`--run-id` still targets a prior run's failures).

**4. Skill-doc polish (from a skill-creator pass over all six touched skills):** thread prompt-mode through eval-analyze/eval-dataset/eval-run guidance, link the orphaned `execution-modes.md`, note `from_traces.py`'s cross-skill ownership, and flag the target artifact for prompt-mode in optimize/review.

Some follow-up that can be done in a future PR: `/eval-optimize` and `/eval-review` still assume a SKILL.md as the edit target. I added a note that for prompt-mode evals the target is the docs/prompt under test, but the deeper rework (actually locate + edit docs vs SKILL.md per mode) is left as future work.

Prashanth684:
@astefanutti thanks! i can squash the commits if necessary and we can probably merge the PR? like you noted any follow up can go in future PRs.

astefanutti:
@Prashanth684 I'm finalizing it, I have a few things to clean up and I'll merge it.

astefanutti (approved):
Thanks @Prashanth684 for all the hard work and iterations!

PR #70, opened 2026-05-21
feat: add dataset.workspace.files for provisioning case source files

StevenTobin:
## Description

Skills that operate on source code (e.g., autofix) need actual files in the workspace, not just input.yaml metadata. Add a dataset.input_files_dir config field (default: files). When a case directory contains a subdirectory with that name, its contents are recursively copied into the workspace root. The directory name is added to _SKIP_NAMES so it is not treated as input data in batch mode. No breaking changes -- cases without a files/ directory are unaffected.

## How Has This Been Tested?

Using autofix-skills as a testbed.

```
Yes, confirmed! The code change is working correctly:

  Dataset structure:
  - Files are in eval/cases/case-001-simple-null-pointer-fix/files/ subdirectory ✓

  Workspace structure:
  - Files copied to workspace root /tmp/agent-eval/2026-05-21-opus-v2/cases/case-001-simple-null-pointer-fix/ ✓
  - user_profile.py and test_user_profile.py are directly accessible ✓

  Execution:
  - Skill is successfully reading the files (logs show "Reading: implement-agent.md") ✓
  - No file-not-found errors ✓

  The updated harness is correctly loading files from the files/ subdirectory (as specified by input_files_dir: files in eval.yaml line 48) and copying them
  to the workspace root where the skill expects them.
```

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

## Summary by CodeRabbit

* **New Features**
  * Dataset settings are grouped under a single dataset section in config.
  * Config can list workspace input files; listed inputs are copied into evaluation workspaces, preserving nested structure.

* **Bug Fixes**
  * Workspace file entries validated (must be relative strings, no parent traversal).
  * Symlink escapes prevented; missing entries are treated as no-ops.

* **Tests**
  * Added comprehensive tests for parsing, copying behavior, and security edge cases.

PR #88, opened 2026-05-31
feat: add execution lifecycle hooks to eval.yaml

stbenjam:
## Summary

Please see `specs/006-execution-hooks/spec.md` for details.  An implementation is included but happy to iterate on the spec to make sure we get the right abstraction here.

Eval cases can depend on external resources (services, databases, fixtures) that today must be managed with ad-hoc scripts outside the harness. This adds lifecycle hooks to `eval.yaml` — user-defined shell commands that run at well-defined points in the eval pipeline, making setup and teardown declarative, version-controlled, and managed by the harness.

Five hook points spanning the full eval lifecycle:

```
  before_all          → start services, extract shared archives
   ├─ before_each     → seed per-case state, extract case-specific data
   │   └─ [skill execution]
   │   └─ after_each  → capture ephemeral state, normalize outputs
   ├─ [collection]
   ├─ before_scoring  → aggregate cross-case data for judges
   └─ after_all       → stop services, clean up (guaranteed, even on failure)
```

  Each hook supports `timeout`, `condition` (skip if shell expression fails), `on_failure` (`fail` or `continue`), and `description` for progress
  output. Hook stdout/stderr is captured to `{run_dir}/hooks/`.

  ## Changes

  - **`agent_eval/config.py`** — `HookEntry` and `HooksConfig` dataclasses, parsed from `hooks:` block in eval.yaml
  - **`agent_eval/hooks.py`** *(new)* — Hook executor (`run_hooks`, `run_hooks_safe`) with timeout enforcement, condition checks, and a standalone CLI
  entry point
  - **`skills/eval-run/scripts/execute.py`** — Calls `before_all`/`after_all` around execution (with `try/finally` for guaranteed cleanup),
  `before_each`/`after_each` around each case
  - **`skills/eval-run/scripts/score.py`** — Calls `before_scoring` hooks at the start of judge scoring
  - **`skills/eval-run/SKILL.md`** — Documents hook behavior, adds `--run-id` to execute.py and `--workspace`/`--model` to score.py invocations
  - **`specs/006-execution-hooks/spec.md`** — Full design spec
  - **`tests/test_hooks.py`** *(new)* — 19 tests covering config parsing, execution, conditions, timeouts, failure modes, env vars, and log capture

  ## Example

  ```yaml
  hooks:
    before_all:
      - command: "docker compose -f evals/services.yaml up -d"
        timeout: 60
        description: "Start eval services"
    before_each:
      - command: "psql -h localhost -U eval -f $CASE_SOURCE_DIR/seed.sql eval_db"
        condition: "test -f $CASE_SOURCE_DIR/seed.sql"
        description: "Seed database for case"
    after_all:
      - command: "docker compose -f evals/services.yaml down -v"
        on_failure: continue
        description: "Tear down services"
```

  Test plan

- [x] All 19 new hook tests pass
- [x] Full test suite (262 tests) passes with no regressions

## Summary by CodeRabbit

* **New Features**
  * Execution lifecycle hooks added (before_all, before_each, after_each, before_scoring, after_all) with conditional execution, per-hook timeouts, fail/continue semantics, log capture, and hook-provided env/data merged and injected into runs and judges.

* **CLI / UX**
  * Added run-id support; score CLI accepts workspace and model context; batch mode warns and ignores per-case hooks; after_all hooks run reliably and hook logs are written per run/case.

* **Documentation**
  * Specs and skill docs updated with hook semantics, env injection, outputs, and file lifecycle.

* **Tests**
  * Extensive unit and integration tests covering parsing, execution, timeouts, outputs, merging, and safety.

### DIFF
--- a/agent_eval/agent/cli_runner.py
+++ b/agent_eval/agent/cli_runner.py
@@ -131,7 +131,7 @@ class CliRunner(EvalRunner):
             with _print_lock:
                 print(f"  {self._log_prefix} | Running: {cmd_str[:120]}", flush=True)
 
-        env = self._build_env()
+        env = self._build_env(settings_path=settings_path)
         start = time.monotonic()
 
         try:
@@ -218,8 +218,12 @@ class CliRunner(EvalRunner):
             return [_sub(part) for part in self._command]
         return _sub(self._command)
 
-    def _build_env(self) -> dict:
-        """Build subprocess environment: inherit env, add extras."""
+    def _build_env(self, settings_path: Optional[Path] = None) -> dict:
+        """Build subprocess environment: inherit env, apply strip list, add extras.
+
+        If settings_path is provided, also merges env vars from the
+        settings.json ``env`` block (written by hook output injection).
+        """
         env = os.environ.copy()
         for k, v in self._extra_env.items():
             if isinstance(v, str) and v.startswith("$"):
@@ -228,6 +232,14 @@ class CliRunner(EvalRunner):
                     env[k] = resolved
             else:
                 env[k] = str(v)
+        # Merge env vars from settings.json (hook outputs, MLflow tracing, etc.)
+        if settings_path and settings_path.exists():
+            try:
+                settings = json.loads(settings_path.read_text())
+                for k, v in (settings.get("env") or {}).items():
+                    env[k] = str(v)
+            except (json.JSONDecodeError, OSError):
+                pass
         return env
 
     @staticmethod
--- a/agent_eval/agent/cli_runner.py
+++ b/agent_eval/agent/cli_runner.py
@@ -87,6 +87,7 @@ class CliRunner(EvalRunner):
         system_prompt: Optional[str] = None,
         max_budget_usd: float = 5.0,
         timeout_s: int = 600,
+        extra_env: Optional[dict] = None,
     ) -> RunResult:
         workspace = workspace.resolve()
         output_dir = workspace / "output"
@@ -131,7 +132,7 @@ class CliRunner(EvalRunner):
             with _print_lock:
                 print(f"  {self._log_prefix} | Running: {cmd_str[:120]}", flush=True)
 
-        env = self._build_env(settings_path=settings_path)
+        env = self._build_env(extra_env=extra_env)
         start = time.monotonic()
 
         try:
@@ -218,12 +219,8 @@ class CliRunner(EvalRunner):
             return [_sub(part) for part in self._command]
         return _sub(self._command)
 
-    def _build_env(self, settings_path: Optional[Path] = None) -> dict:
-        """Build subprocess environment: inherit env, apply strip list, add extras.
-
-        If settings_path is provided, also merges env vars from the
-        settings.json ``env`` block (written by hook output injection).
-        """
+    def _build_env(self, extra_env=None) -> dict:
+        """Build subprocess environment: inherit env, add extras."""
         env = os.environ.copy()
         for k, v in self._extra_env.items():
             if isinstance(v, str) and v.startswith("$"):
@@ -232,14 +229,9 @@ class CliRunner(EvalRunner):
                     env[k] = resolved
             else:
                 env[k] = str(v)
-        # Merge env vars from settings.json (hook outputs, MLflow tracing, etc.)
-        if settings_path and settings_path.exists():
-            try:
-                settings = json.loads(settings_path.read_text())
-                for k, v in (settings.get("env") or {}).items():
-                    env[k] = str(v)
-            except (json.JSONDecodeError, OSError):
-                pass
+        if extra_env:
+            for k, v in extra_env.items():
+                env[k] = str(v)
         return env
 
     @staticmethod
--- a/agent_eval/config.py
+++ b/agent_eval/config.py
@@ -543,6 +543,20 @@ class EvalConfig:
                 ))
             setattr(config.hooks, phase, entries)
 
+        if config.execution.mode == "batch":
+            per_case = []
+            if config.hooks.before_each:
+                per_case.append("before_each")
+            if config.hooks.after_each:
+                per_case.append("after_each")
+            if per_case:
+                import warnings
+                warnings.warn(
+                    f"hooks.{', '.join(per_case)} ignored in batch mode "
+                    f"(per-case hooks only run in case/prompt mode)",
+                    stacklevel=2,
+                )
+
         if config.skill and not _is_valid_eval_name(config.skill):
             raise ValueError(
                 f"Invalid skill name in {path}: {config.skill!r}")
--- a/agent_eval/config.py
+++ b/agent_eval/config.py
@@ -524,11 +524,21 @@ class EvalConfig:
         for phase in phases:
             entries = []
             for h in (hooks_raw.get(phase) or []):
+                on_failure_val = h.get("on_failure", "fail")
+                if on_failure_val not in ("fail", "continue"):
+                    raise ValueError(
+                        f"hooks.{phase}: on_failure must be 'fail' or "
+                        f"'continue', got '{on_failure_val}'")
+                timeout_val = h.get("timeout", 120)
+                if not isinstance(timeout_val, int) or timeout_val <= 0:
+                    raise ValueError(
+                        f"hooks.{phase}: timeout must be a positive "
+                        f"integer, got {timeout_val}")
                 entries.append(HookEntry(
                     command=h.get("command", ""),
-                    timeout=h.get("timeout", 120),
+                    timeout=timeout_val,
                     description=h.get("description", ""),
-                    on_failure=h.get("on_failure", "fail"),
+                    on_failure=on_failure_val,
                     condition=h.get("condition", ""),
                 ))
             setattr(config.hooks, phase, entries)
--- a/agent_eval/config.py
+++ b/agent_eval/config.py
@@ -125,6 +125,26 @@ class InputsConfig:
 
 
 @dataclass
+class HookEntry:
+    """A single lifecycle hook command."""
+    command: str = ""
+    timeout: int = 120
+    description: str = ""
+    on_failure: str = "fail"  # "fail" | "continue"
+    condition: str = ""
+
+
+@dataclass
+class HooksConfig:
+    """Lifecycle hooks that run at defined points in the eval pipeline."""
+    before_all: list = field(default_factory=list)
+    before_each: list = field(default_factory=list)
+    after_each: list = field(default_factory=list)
+    before_scoring: list = field(default_factory=list)
+    after_all: list = field(default_factory=list)
+
+
+@dataclass
 class ExecutionConfig:
     """How the skill is invoked against test cases.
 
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

**Comment by ralphbean** [comment]
This is exciting — we've been working on functional tests for fullsend agents (fullsend-ai/fullsend#1682) and ran into exactly the problem this solves. We have a use case that would benefit from one additional capability though, and I'm curious if it resonates with what you're thinking here.

Our functional tests need to create ephemeral GitHub fixtures (repos, issues, PRs) before each case runs, then point the skill at the right fixture. The `before_each` hook is a natural place to do that setup — but the fixture URL that the setup script produces needs to flow into the skill execution as an env var (e.g. `GITHUB_ISSUE_URL`). Right now there's no channel for that. The hook runs, the skill runs, but they can't pass dynamic state between them.

Here's a proposal: what if `before_each` hooks could write a file to a known path — say `$CASE_WORKSPACE/.hook-env` — and the harness reads it after the hooks complete, injecting those key-value pairs into the skill's environment? Something like:

```bash
# in a before_each hook
echo "GITHUB_ISSUE_URL=https://github.com/org/repo/issues/2" >> "$CASE_WORKSPACE/.hook-env"
echo "EPHEMERAL_REPO=org/eval-repo-abc123" >> "$CASE_WORKSPACE/.hook-env"
```

The harness picks those up and merges them into the `settings.json` env block (or the CLI runner's env) before skill execution starts. Same idea could work for `before_all` writing to `$AGENT_EVAL_WORKSPACE/.hook-env` for vars that apply to all cases.

With that in place, our entire functional test setup collapses to hooks — no custom runner script at all:

```yaml
hooks:
  before_each:
    - command: "$CASE_SOURCE_DIR/setup.sh $CASE_WORKSPACE"
      description: "Create ephemeral repo and fixtures"
  after_each:
    - command: "$CASE_SOURCE_DIR/capture.sh $CASE_WORKSPACE"
      description: "Capture fixture state"
    - command: "eval/scripts/teardown-repo.sh $CASE_WORKSPACE"
      on_failure: continue
      description: "Delete ephemeral repo"
```

I think there's a broader pattern here beyond our use case — any eval where the thing-to-evaluate is dynamically provisioned (a service URL, a database connection string, a temporary API key) would benefit from hooks being able to feed state forward into skill execution.

ralphbean:
This is great, b109284 addresses my earlier question about dynamic state flow almost exactly. The `.hook-outputs.yaml` approach is better than the `.hook-env` dotenv idea I floated — handles quoting/multiline cleanly and the `data` channel for judges is a nice bonus.

Two things I noticed:

**1. `after_each` not in try/finally**

`after_all` is wrapped in try/finally at the outer loop level, which guarantees cleanup. But `after_each` in `_run_single_case` runs sequentially after `runner.run_skill()` with no try/finally. I know both runners are designed to always return a `RunResult`, so this probably never bites in practice — but given the outer loop already sets the precedent, would it make sense to do the same for symmetry? Our use case will have `before_each` creating ephemeral GitHub repos and `after_each` deleting them, so if anything exotic does raise, leaked repos would be annoying to track down.

**2. Hook output env vars and the CLI runner**

`inject_hook_env` writes to `.claude/settings.json`, which Claude Code picks up via `--settings`. But the CLI runner's `_build_env()` only reads from `os.environ` and `self._extra_env` — it never looks at `settings.json`. The spec mentions "For CLI runners, merged into the subprocess env dict" but I don't think the implementation does that yet.

This one affects us directly — we use `runner.type: cli` and want hook outputs to inject env vars into the runner subprocess. Would it make sense to have `inject_hook_env` also merge into the process environment (or have `_build_env()` read from settings.json), so both runner types pick up hook-output vars?

ralphbean (approved):

skills/eval-run/scripts/execute.py
```diff
@@ -277,16 +315,86 @@ def _run_single_case(runner, skill_name, case_id, case_ws, output_dir,
...
+    # Per-case hooks: build case-specific env, run before_each/after_each.
+    # The entire block is wrapped in try/finally so that after_each cleanup
+    # hooks fire even if before_each or run_skill raises.
+    case_hook_env = None
+    merged_hook_data = {}
+    result = None
+    error_msg = ""
+    try:
+        if config and config.hooks and hook_env:
+            dataset_path = config.dataset_path
```
astefanutti: This branch predates #70, which removed `EvalConfig.dataset_path` (now nested as `config.dataset.path`). #88 still reads the old attribute in `execute.py`:

```python
dataset_path = config.dataset_path   # removed by #70 → AttributeError
case_source_dir = str(Path(dataset_path).resolve() / case_id) if dataset_path else ""
```

Could you please:

1. Rebase onto `main` (with #70).
2. `config.dataset_path` → `config.dataset.path` in `execute.py` (and anywhere else).
3. Update the test helper to use the new `DatasetConfig` structure instead of setting `dataset_path` directly.

specs/006-execution-hooks/spec.md
```diff
@@ -0,0 +1,518 @@
...
+| `AGENT_EVAL_RUN_ID` | all | Current run ID (e.g., `2026-05-31-sonnet`) |
+| `AGENT_EVAL_CONFIG` | all | Absolute path to eval.yaml |
+| `AGENT_EVAL_PROJECT_ROOT` | all | Project root directory |
+| `AGENT_EVAL_MODEL` | all | Skill model being tested (e.g., `sonnet`) |
+| `CASE_ID` | per-case hooks | Case ID (e.g., `case-001-name`) |
+| `CASE_WORKSPACE` | per-case hooks | Absolute path to case workspace directory |
+| `CASE_SOURCE_DIR` | per-case hooks | Absolute path to the original case directory in the dataset |
+| `CASE_INPUT` | per-case hooks | Absolute path to `input.yaml` in the case workspace |
+
+`CASE_SOURCE_DIR` is the key variable — it gives hooks access to files in the dataset case directory (like `snapshot.tar.gz`, `fixtures/`, test data) that workspace.py deliberately does not copy into the case workspace to avoid leaking annotations and gold standards to the skill.
```
astefanutti: #70 (merged) added `dataset.workspace.files` that's a declarative way to copy case files into the workspace. So workspace.py no longer "deliberately does not copy" case files; it copies whatever's whitelisted there.

The two mechanisms are complementary and compose cleanly (workspace files land during setup, `before_each` runs after), so I'd just reconcile the framing here:

- **`dataset.workspace.files`** (#70) — declarative copy of static case files into the workspace.
- **`before_each` + `CASE_SOURCE_DIR`** — the imperative escape hatch for what a static copy can't do: extracting archives (`snapshot.tar.gz`), seeding a DB, pulling volumes, computed/dynamic provisioning.

agent_eval/hooks.py
```diff
@@ -0,0 +1,346 @@
...
+                content = {}
+            try:
+                path.unlink()
+            except OSError:
+                pass
+            return content
+    return {}
+
+
+def inject_hook_env(workspace_path, env_vars):
```
astefanutti: This is a second env-injection pipeline parallel to `execution.env`, and I think hook env should ride that existing rail instead.

`execution.env` already injects env into the skill, with clean per-runner paths: settings.json `env` block for claude-code, `_build_env()` for the cli runner. `inject_hook_env` reimplements that — it writes settings.json, and then cli_runner had to add a settings.json round-trip in `_build_env` (cli_runner.py:235) to read it back, even though its normal env channel is `_build_env` straight from config. That duplication is what makes the two runners feel asymmetric and adds a path to keep in sync.

Conceptually a hook emitting `env: {DATABASE_URL: ...}` is just a **dynamically-computed `execution.env`** — same thing as a static `execution.env: {DATABASE_URL: ...}`, only the value is produced at runtime. So it should go through the same mechanism:

- After `collect_hook_outputs(...)`, merge the hook `env` into the same env dict that `execution.env` is injected from, and let the existing per-runner injection deliver it.
- Drop `inject_hook_env` and the bespoke settings.json read in `cli_runner._build_env`.

Benefits:
- Per-runner-correct for free (goes wherever `execution.env` already goes), so no claude-code/cli asymmetry.
- One precedence chain, documented once: `allowlist < runner.env (#108) < execution.env < hook env` (dynamic overrides static), identical on every runner.
- No parallel code path, and it composes cleanly with #108's `runner.env`.

tl;dr: I'd route hook `env` through `execution.env`'s existing one and avoid building a second env pipeline.

agent_eval/agent/cli_runner.py
```diff
@@ -228,6 +232,14 @@ def _build_env(self) -> dict:
                     env[k] = resolved
             else:
                 env[k] = str(v)
+        # Merge env vars from settings.json (hook outputs, MLflow tracing, etc.)
```
astefanutti: This settings.json round-trip is the symptom of the parallel env pipeline (see my note on `hooks.py` `inject_hook_env`). cli_runner's normal env channel is `_build_env()` straight from config (that's how `execution.env` reaches it) — reading `.claude/settings.json` back here only exists to pick up `inject_hook_env`'s writes.

If hook `env` is funneled through `execution.env`'s existing injection instead, this block goes away entirely: the hook values arrive via the same path as `execution.env`, with no settings.json detour and no claude-code/cli asymmetry.

agent_eval/config.py
```diff
@@ -429,6 +452,32 @@ def from_yaml(cls, path: str | Path) -> "EvalConfig":
...
+                        f"hooks.{phase}: timeout must be a positive "
+                        f"integer, got {timeout_val}")
+                entries.append(HookEntry(
+                    command=h.get("command", ""),
+                    timeout=timeout_val,
+                    description=h.get("description", ""),
+                    on_failure=on_failure_val,
+                    condition=h.get("condition", ""),
+                ))
+            setattr(config.hooks, phase, entries)
```
astefanutti, line 544: The hooks block parses all five phases unconditionally, with no cross-check against `execution.mode` — but `before_each`/`after_each` only run in `case`/`prompt` mode (via `_execute_per_case` → `_run_single_case`). In **batch** mode the single-invocation path runs only `before_all`/`after_all`, so per-case hooks are **silently ignored**: a batch eval with `before_each:` configured parses fine, runs, and never executes them — no error, no warning.

This is the same class of issue as #111 ("workspace.files should warn or error in batch mode"), where the team chose to warn rather than silently drop. Per-case hooks should follow that precedent.

Suggest a load-time check here: if `execution.mode == "batch"` and `before_each`/`after_each` are non-empty, warn (or error) that per-case hooks don't run in batch mode. Validating at config load fails loud before any work runs, and `config.execution.mode` is already available at this point.

While here, worth documenting in the spec that batch mode supports only `before_all`/`after_all`/`before_scoring`, and that the per-case env vars (`CASE_SOURCE_DIR`, `CASE_WORKSPACE`, `CASE_INPUT`) are case/prompt-mode only.

astefanutti (commented):
@stbenjam thanks, I've left a couple of comments and it'll be good to go.

astefanutti (approved):
Thanks @stbenjam, LGTM!

ralphbean:
Thanks for this!

PR #98, opened 2026-06-02
fix(score): use final result text as conversation content for judges

### LINKED ISSUE #94: score.py: conversation key contains raw tool calls instead of final skill output
## Problem

`load_case_record()` builds the `conversation` key by concatenating all
assistant turns — including tool calls, reasoning, and intermediate outputs.
LLM judges receive this bloated context instead of the skill's actual final output.

## Root cause

`extract_conversation_text()` in `score.py` joins every assistant turn.
The `result` event carries the final output text in `result.text` but it is
never extracted or surfaced to judges.

## Impact

Judges consistently score tool call chains rather than skill output, producing
unreliable scores for any skill that uses multiple tool calls before answering.
Related to #63 — even when `{{ conversation }}` is used correctly in judge
prompts, the content it receives is wrong.

## Reproduction

Run any multi-tool skill evaluation. Inspect the `conversation` value passed
to a judge — it will contain tool_use blocks and reasoning text rather than
the final skill response.

## Suggested fix

Extract `text` from the `result` event in `events.py`:

```python
def _parse_result_event(obj):
    return {
        "type": "result",
        "text": obj.get("result", ""),  # final skill output
        "cost_usd": obj.get("total_cost_usd"),
    }
```

In `score.py`, prefer `result.text` over `extract_conversation_text()` when present:

```python
result_events = [e for e in events if e.get("type") == "result" and e.get("text")]
if result_events:
    record["conversation"] = result_events[-1]["text"]
else:
    record["conversation"] = extract_conversation_text(events)
```

## Workaround

None reliable — manually inspect `stdout.log` to verify what judges are scoring.

abhijeet-dhumal:
Fixes #94

## Problem
Judges were receiving a concatenation of all assistant turns (tool calls, reasoning, intermediate outputs) instead of the final skill output.

## Changes
- `events.py`: extract `text` from the result event (`result.text`)
- `score.py`: prefer `result.text` over `extract_conversation_text()` when present

## Testing
Verified with sprint-health-check, sprint-planning, and sprint-retrospective skills — judges now score the final output text rather than intermediate tool calls.

hc-07-span-detail

## Summary by CodeRabbit

* **Improvements**
  * Result events now include final skill output text for better visibility into evaluation results.
  * Enhanced conversation context extraction from event streams by prioritizing recent skill outputs.
  * Increased token budget for LLM judge responses to enable more detailed evaluations.
  * Added automatic retry mechanism for LLM responses that reach token limits.

### DIFF
--- a/agent_eval/events.py
+++ b/agent_eval/events.py
@@ -161,6 +161,7 @@ def _parse_user_tool_results(obj, tool_id_to_name, result_cap):
 def _parse_result_event(obj):
     return {
         "type": "result",
+        "text": obj.get("result", ""),   # final skill output text
         "cost_usd": obj.get("total_cost_usd"),
         "num_turns": obj.get("num_turns"),
         "timestamp": None,
--- a/skills/eval-run/scripts/score.py
+++ b/skills/eval-run/scripts/score.py
@@ -196,9 +196,17 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
         record["events"] = []
 
     # --- Conversation text (convenience key for check judges) ---
+    # Prefer the result event's final output text over concatenating all
+    # intermediate assistant turns, which bloats the context with tool calls
+    # and reasoning text.
     if record["events"]:
         from agent_eval.events import extract_conversation_text
-        record["conversation"] = extract_conversation_text(record["events"])
+        result_events = [e for e in record["events"]
+                         if e.get("type") == "result" and e.get("text")]
+        if result_events:
+            record["conversation"] = result_events[-1]["text"]
+        else:
+            record["conversation"] = extract_conversation_text(record["events"])
     else:
         record["conversation"] = ""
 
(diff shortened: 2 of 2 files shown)


abhijeet-dhumal:
## Validation

Tested across 3 skills: `sprint-health-check`, `sprint-planning`, `sprint-retrospective` (5 cases each via agent-eval-harness).

**Before fix:** Judges received a concatenation of all assistant turns — raw tool call blocks, Jira API responses, and reasoning text (~5000 tokens of noise before the actual skill output).

**After fix:** Judges receive only the final output text from the `result` event — the actual markdown report the skill produced.

**Observed impact on judge scores:**

| Judge | Before (avg) | After (avg) |
|---|---|---|
| `capacity_table_quality` | 1.7 / 5 | 4.8 / 5 |
| `completion_percentage_accuracy` | 2.1 / 5 | 5.0 / 5 |
| `burndown_verdict_accuracy` | 2.3 / 5 | 4.75 / 5 |
| `cycle_time_insight` | 1.5 / 5 | 4.63 / 5 |

Scores were unreliable before because judges were evaluating tool call chains rather than skill output. After fix, scores align with manual review of the outputs.

PR #99, opened 2026-06-02
fix(trace_builder): return server trace ID from _log_trace to fix FK failures

### LINKED ISSUE #95: trace_builder.py: _log_trace server ID discarded causes FOREIGN KEY failures in log_feedback
## Problem

`MlflowClient._log_trace()` returns a backend-generated trace ID that differs
from the client-side ID in `trace_dict["info"]["trace_id"]`. The return value
was discarded, so downstream `log_feedback` calls reference an ID the server
does not recognise.

## Impact

All judge assessments fail with `FOREIGN KEY constraint failed`.
The MLflow Quality tab shows no scores for any eval run.

## Reproduction

1. Run `eval-mlflow` with LLM judges configured.
2. Check stderr for `FOREIGN KEY constraint failed` errors.
3. Confirm the MLflow Quality tab is empty despite judges running.

## Suggested fix

```python
# Before
client._log_trace(trace)
return trace_dict["info"]["trace_id"]

# After — return the server-generated ID for downstream FK operations
server_trace_id = client._log_trace(trace)
return server_trace_id or trace_dict["info"]["trace_id"]
```

## Workaround

None — assessments cannot be logged without this fix.

abhijeet-dhumal:
Fixes #95

## Problem
`MlflowClient._log_trace()` returns a backend-generated trace ID that differs from the client-side ID. The return value was discarded, so `log_feedback` referenced an unknown ID and failed with `FOREIGN KEY constraint failed`.

## Change
Capture and return the server-generated trace ID:
```python
server_trace_id = client._log_trace(trace)
return server_trace_id or trace_dict["info"]["trace_id"]
```

## Testing
Verified that judge assessments now appear in the MLflow Quality tab after this fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Trace ID retrieval now returns the backend-generated identifier when available.
  * If the backend fails to produce an ID, a warning is emitted and the original trace ID is used as a fallback.

### DIFF
--- a/agent_eval/mlflow/trace_builder.py
+++ b/agent_eval/mlflow/trace_builder.py
@@ -857,8 +857,11 @@ def log_trace(trace_dict):
         client = MlflowClient()
         # No public API for logging pre-built traces as of MLflow 3.5.
         # _log_trace is the only way to submit a Trace object.
-        client._log_trace(trace)
-        return trace_dict["info"]["trace_id"]
+        # IMPORTANT: _log_trace returns a backend-generated trace ID that
+        # differs from the client-provided trace_dict["info"]["trace_id"].
+        # Always return the backend ID so downstream FK operations succeed.
+        server_trace_id = client._log_trace(trace)
+        return server_trace_id or trace_dict["info"]["trace_id"]
     except Exception as e:
         print(f"WARNING: failed to log trace: {e}", file=sys.stderr)
         return None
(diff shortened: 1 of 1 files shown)


abhijeet-dhumal:
## Validation

Tested with `sprint-health-check` skill (9 judges, 5 cases) using MLflow 3.x local tracking server.

**Before fix:** All `log_feedback` calls failed silently with `FOREIGN KEY constraint failed`. The MLflow Quality tab showed 0 assessments despite judges running successfully.

**After fix:** All judge assessments log correctly. MLflow Quality tab shows:
- 9 judges × 5 cases = 45 assessments visible
- `burndown_verdict_accuracy` avg 4.75, `grace_period_logic` avg 5.0
- Assessment rationales visible inline (up to 2000 chars)

**Root cause confirmed:** Added debug logging to print both IDs — the client-side ID (e.g. `tr-abc123`) and the server-returned ID (e.g. `tr-xyz789`) — and verified they differ. `log_feedback` was referencing the client-side ID which the server did not recognise.

astefanutti (approved):
Thanks @abhijeet-dhumal!

PR #121, opened 2026-06-10
fix: prevent silent workspace deletion on re-run

not-stbenjam:
## Summary

- `workspace.py` previously called `shutil.rmtree()` unconditionally when a workspace directory already existed, silently destroying all uncollected artifacts (transcripts, subagent logs, output files)
- This caused data loss when retrying failed cases with the same run-id — the workspace for previously successful cases was wiped before `collect.py` had a chance to harvest their outputs
- Now requires `--force` to overwrite an existing workspace, otherwise exits with a clear error message suggesting `--force` or a different `--run-id`

## Test plan

- [ ] Run `workspace.py` with a run-id that already exists — should error with guidance
- [ ] Run `workspace.py --force` with existing run-id — should warn and proceed
- [ ] Run `workspace.py` with a fresh run-id — should work as before (no behavior change)

## Summary by CodeRabbit

* **New Features**
  * Added a `--force` option to overwrite an existing workspace when rerunning the same job ID.
  * Improved workspace creation behavior by using a deterministic location for forced runs and a separate temporary workspace for normal runs.
  * Tightened workspace setup permissions for better isolation and safety.


jeremyeder:
lgtm

rhuss:
Good fix for a real data loss scenario. The `mkdtemp` approach is the right default, and keeping `--force` for deterministic paths is a clean escape hatch. The workspace path flows via stdout (`WORKSPACE: <path>`) so downstream scripts pick it up regardless.

Two things before merging:

1. **Rebase needed.** The branch is 65 commits behind main. Notably, #128 landed `_validate_path_segment()` in `config.py` and #70 added `workspace.files` support. Both touch workspace.py so there will be conflicts.

2. **Dropped path traversal check.** The current main has an explicit parent-directory guard:
   ```python
   if base_dir not in workspace.parents and workspace != base_dir:
       print("ERROR: invalid run-id path")
   ```
   This was removed in the PR. The regex validation earlier (`[A-Za-z0-9._-]+`) prevents traversal in practice, but removing an explicit safety check is not ideal. After rebasing, please pick up `_validate_path_segment()` from #128 and use it to validate `args.run_id` in the `--force` path. That way the defense-in-depth is preserved (and improved).

not-stbenjam:
Thanks for the review! Both items addressed:

1. **Rebased** — merged `main` into the branch (65 commits caught up). Auto-merge resolved `workspace.py` cleanly.

2. **Path traversal guard restored** — added `_validate_path_segment()` from #128 to validate `args.run_id` in the `--force` path before it's used to construct the deterministic workspace directory (which gets `shutil.rmtree`'d). The regex check at line 72 still provides first-pass validation, and `_validate_path_segment` adds defense-in-depth with explicit checks for path separators, `..`/`.`, and control characters.

not-stbenjam:
PR is stale and I think this behavior has been fixed; haven't seen it.

PR #125, opened 2026-06-13
Warn on stale inline frontmatter fields

### LINKED ISSUE #33: Detect stale judge field references before scoring
### Problem
                                                                                                                                                                                                                                                                                          

When a skill changes its output schema (e.g., renaming a frontmatter field from strat_key to source_key), inline check judges that reference the old field name silently fail with misleading rationale. The judge reports "bad strat_key" — suggesting the skill produced bad output, when in reality the field was renamed and the judge is stale.
                                                                                                                                                                                                                                                                                          
Discovered during evaluation of test-plan.create where the frontmatter_valid judge checked fm.get("strat_key") but the skill now writes source_key. All 10 cases showed 0% pass rate for a judge that should have been 100%.

#### Impact                                                                                                                                                                                                                                                                                  
                           
  - Wasted scoring cost — LLM judges ran against all 10 cases before the issue was noticed
  - Misleading results — the 0% pass rate looked like a skill regression, not a config issue
  - Silent failure — no warning that the judge was accessing fields that don't exist in the output                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                        
#### Proposed Solution                                                                                                                                                                                                                                                                       

                                                                                                                                                                                                                                                                                          
- Option 1 (recommended): Pre-score field validation in score.py — Before running judges, scan inline check snippets for fm.get("field_name") patterns. Read one collected artifact and extract available field names. If a judge references a field that doesn't exist, warn before scoring.
                                                                                                                                                                                                                                                                                          
- Option 2: score.py --dry-run — Run inline check judges against one case and report field access vs availability, without scoring.                                                                                                                                                       
  
- Option 3: Schema drift detection in /eval-analyze --update — Diff field names in existing judges against the skill's current output schema.                                                                                                                                             
                           
#### Context                                                                                                                                                                                                                                                                                 
                           
  - Run 2026-04-28-opus, skill test-plan.create                                                                                                                                                                                                                                           
  - Skill's frontmatter.py renamed strat_key → source_key
  - eval.yaml judge still referenced strat_key                                                                                                                                                                                                                                            
  - All other judges passed correctly

run against the test-case.create skill from - 

#### Run Details

Run Overview: 2026-04-28-opus                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
                           
  - 10/10 cases passed execution (all produced 4 artifacts)                                                                                                                                                                                                                               
  - 0/10 cases pass frontmatter_valid — this is a known judge bug (source_key vs strat_key field name mismatch)
  - Grounding scores range from 2-5, with 4 cases depressed by a template rendering issue in the LLM judge                                                                                                                                                                                
  - Specificity is perfect across all 10 cases

[2026-04-28-opus.zip](https://github.com/user-attachments/files/27185127/2026-04-28-opus.zip)

luohui1:
## Summary

- add a warning-only preflight for inline check judges that reference YAML frontmatter fields via `fm.get(...)`
- infer the output artifact that feeds the `fm` assignment so unrelated `_content` reads do not create noisy warnings
- add regression tests for stale fields, empty frontmatter, loader-error tolerance, comments, and multi-artifact cases

Closes #33.

## Validation

- `.\.venv\Scripts\python.exe -m pytest tests/test_score_builtin.py::TestInlineJudgeFieldValidation -q`
- `.\.venv\Scripts\python.exe -m pytest tests/test_score_builtin.py -q`
- `git diff --cached --check`
- `UV_CACHE_DIR=C:\tmp\uv-cache uvx skillsaw`

Full `.\.venv\Scripts\python.exe -m pytest tests/ -v` previously failed in my Windows environment with existing shell/symlink/path-related failures: 46 failed, 323 passed, 2 deselected.

## Summary by CodeRabbit

* **New Features**
  * Inline judge checks now detect references to missing or outdated frontmatter fields and issue warnings, helping identify configuration problems earlier.
  * Validation distinguishes between absent, empty, and unreadable frontmatter and continues scoring available cases when individual checks encounter errors.

* **Tests**
  * Added comprehensive coverage for field validation, commented references, output-specific behavior, skipped and passing judges, loading failures, and scoring resilience.

### DIFF
--- a/skills/eval-run/scripts/score.py
+++ b/skills/eval-run/scripts/score.py
@@ -1252,47 +1252,24 @@ def _parse_inline_check_source(source):
 # Names an inline check conventionally binds the parsed frontmatter to. The
 # analysis is name-based, so anything else is simply not analysed — silence,
 # never a wrong warning.
-_FRONTMATTER_NAMES = {"fm", "frontmatter", "meta"}
+_FRONTMATTER_NAMES = {"fm", "frontmatter"}
 
 
 def _extract_frontmatter_field_refs(source):
-    """Fields an inline check REQUIRES from the frontmatter.
+    """Frontmatter field names an inline check mentions.
 
-    Three idioms count as a requirement — `fm.get("x")`, `fm["x"]`, and
-    `"x" in fm` / `"x" not in fm`, the last being the one the README, the
-    cookbook and the eval.yaml template all use.
-
-    Two do not, and skipping them is what keeps this quiet on healthy configs:
-    `fm.get("x", default)` states outright that the field is optional, and a
-    reference whose only use is a truth test (`if fm.get("x"): ...`) is an
-    absence assertion — it warns precisely when the skill is behaving.
+    Every literal reference counts — `fm.get("x")`, `fm.get("x", default)`,
+    `fm["x"]`, `"x" in fm`. Deliberately no attempt to infer whether the field
+    is *required*: intent is not in the syntax. `if fm.get("x"): return True`
+    and `if fm.get("x"): return False` are the same expression and opposite
+    requirements, and the project's own template tells authors to pass a
+    default to every lookup, so a default says nothing either. Precision comes
+    from only reporting judges that actually failed (see `score_cases`), not
+    from guessing here.
     """
     tree = _parse_inline_check_source(source)
     if tree is None:
         return []
-
-    # Polarity decides intent. `if fm.get("x")` asserts the field should be
-    # ABSENT — warning there fires when the skill is behaving. `if not
-    # fm.get("x")` is the opposite: it requires the field, and is the exact
-    # shape issue #33 is about. So collect POSITIVE truth tests only, and step
-    # over a `not` rather than through it.
-    optional = set()
-
-    def _mark_positive_tests(node):
-        if isinstance(node, ast.UnaryOp) and isinstance(node.op, ast.Not):
-            return                      # negated: a requirement, keep it
-        if isinstance(node, ast.BoolOp):
-            for value in node.values:
-                _mark_positive_tests(value)
-            return
-        optional.add(id(node))
-
-    for node in ast.walk(tree):
-        for attr in ("test",):
-            test = getattr(node, attr, None)
-            if test is not None:
-                _mark_positive_tests(test)
-
     refs = set()
     for node in ast.walk(tree):
         name = None
@@ -1301,17 +1278,13 @@ def _extract_frontmatter_field_refs(source):
                 and node.func.attr == "get"
                 and _is_frontmatter_name(node.func.value)
                 and node.args):
-            if len(node.args) > 1:      # an explicit default = optional
-                continue
-            if id(node) in optional:    # `if fm.get("x")` = expects absence
-                continue
             name = _literal_string(node.args[0])
         elif (isinstance(node, ast.Subscript)
-              and _is_frontmatter_name(node.value)):
+              and _is_frontmatter_name(node.value)
+              and isinstance(node.ctx, ast.Load)):   # not a write or a del
             name = _literal_string(node.slice)
         elif isinstance(node, ast.Compare) and len(node.ops) == 1:
-            op = node.ops[0]
-            if (isinstance(op, (ast.In, ast.NotIn))
+            if (isinstance(node.ops[0], (ast.In, ast.NotIn))
                     and _is_frontmatter_name(node.comparators[0])):
                 name = _literal_string(node.left)
         if name:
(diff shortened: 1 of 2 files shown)


astefanutti:
Sorry this sat for two months — the problem you picked is real and the diagnosis in #33 is right. I've rebased onto `main` (it was 389 commits behind, so CI was running against a June tree) and pushed four commits rather than sending you a list to work through. Your commit is preserved as `8ecf792`, just reparented. Happy to drop or rework any of it.

## The one that had to change: it could abort a run

`_extract_yaml_frontmatter_keys` caught `yaml.YAMLError`, but `yaml.safe_load` raises a bare `ValueError` for an out-of-range date, and the existing `try/except` covers only `load_case_record`:

```
two cases, case-001 frontmatter has `due: 2026-02-30`
  main:   COMPLETED — case-002 True, case-001 records a per-case error
  before: ABORTED   — ValueError out of score_cases, whole run lost
```

An LLM emitting Feb 30 is mundane, and the probe runs *before* any judge — after the agent has already been paid for. So it destroyed exactly the evidence it exists to explain. `0ebb6fb` catches `Exception`, wraps the whole probe, and distinguishes *unreadable* frontmatter (unknown → stay quiet) from *absent* frontmatter (fields really are missing → still warns, as your `test_referenced_artifact_without_frontmatter_warns` asks).

## The bigger one: I tried your approach harder and it doesn't hold

I first tried to sharpen the `fm.get()` matcher — skip an explicit default, skip a positive truth test. An adversarial pass showed it backwards in exactly the cases that matter:

```
if fm.get("k"): return True     # requires it
if fm.get("k"): return False    # requires it GONE
```

Same expression, opposite meanings — intent isn't in the syntax. Worse, `eval-yaml-template.md:554` tells authors to *"use `.get()` with defaults for all dict lookups"*, so treating a default as "optional" silenced the feature on the style the project recommends, including #33's own judge written that way. Meanwhile `if "legacy" in fm: return False` warned on a run passing 100%.

So `fe26601` stops guessing. Every literal reference is collected, and the warning is gated on **the judge's own verdict**: it fires only for a check judge that returned False on every case, or errored on every case. A passing judge is never reported, so the false-positive class disappears by construction rather than by heuristic.

That also happens to be #33's symptom — a 0% pass rate that looked like a skill regression — so the message now says so.

Two consequences worth flagging, both arguably wins:

- **The probe moved after scoring.** The pre-scoring position never actually saved anything: it printed and fell through, and every LLM judge still ran. Nothing was being avoided.
- **`_judge_runs_on` is gone**, because a judge skipped by its `if:` has no verdict to gate on. That was one of the review items, solved for free.

Also in `faf1954`: it read only `case_dirs[0]`, and a case that produced nothing is a normal outcome, so one bad case spoke for the whole run. It now reads up to three and warns only when a field is absent from all of them.

## Smaller

`meta` is no longer treated as a frontmatter alias — it's just as conventionally a parsed JSON blob, which has no frontmatter to be missing from, and it was warning on those. Subscript writes and `del fm["k"]` no longer count as reads. `"x" in fm` / `fm["x"]` now do.

## Tests

Four mutations that previously survived now fail: the probe budget, unreadable-vs-absent frontmatter, the evidence gate, and the bare-`ValueError` catch. Your six original tests all still pass unchanged.

1201 passing, 12 pre-existing environment failures (missing `anova` extra, responses-API integration) — same set as `main`.

## Still open

No per-judge opt-out. If a judge does trip a false positive there's no `warn: false`, and the probe's frontmatter parser and the judges' own parsers are two different parsers that disagree on edge cases like a `---` inside a value. Neither blocks this, but both are worth a follow-up.

astefanutti:
Good catch, and worse than a nit — fixed in `c07ca2a`.

`split("---", 2)` isn't line-anchored, so a `---` inside a scalar truncates the block:

```
---
title: foo --- bar
status: ok
---
```

`parts[1]` is `"\ntitle: foo "`, and `status` reads as absent from an artifact that plainly has it. Agent-written markdown controls that text, so it's reachable on an ordinary run, and it produces the worst kind of warning — one naming a field the author can see in the file.

It compounds with the evidence gate too: the judges parse the same way, so the judge also fails, the gate opens, and the diagnostic then explains a real failure with the wrong cause.

Anchored with a regex as suggested. Verified across the shapes that matter beyond the one reported — CRLF, a block ending at EOF with no trailing newline, a longer `----` fence, unclosed (still `None`), and no block at all (still an empty set). Reverting to the plain split fails the new test.

1202 passing, same 12 pre-existing environment failures.

On the Ruff `BLE001` note about the blind `except Exception` at 1396 — that one is deliberate and I'd keep it. The whole point of `0ebb6fb` is that this probe runs after the agent has been paid for and must never be able to take a run down; `yaml.safe_load` alone raises `YAMLError`, bare `ValueError`, and `RecursionError` depending on input, so enumerating them is how the original bug happened.

astefanutti (approved):
Thanks @luohui1!

PR #126, opened 2026-06-15
refactor: Feat/harbor integration

abhijeet-dhumal:
## Description

Three OpenShift-specific reliability fixes for `KubernetesEnvironment`, plus two usability improvements.

#### HAProxy idle-connection WebSocket deadlock (`_ws_exec`)
OpenShift's HAProxy router silently drops WebSocket connections idle for ≥60 s, killing long-running agent execs mid-stream with `WebSocketConnectionClosedException`.  
The original `_ws_exec` also deadlocks on `resp.close()` / `read_channel()` when the connection is already gone.

**Fix:** launch the agent command in the background inside the pod (writing stdout/stderr to `/tmp` files), then poll with a series of short-lived WebSocket calls (`_ws_exec_short`) each completing in <1 s — well below the HAProxy idle threshold. A `threading.Thread` with a hard wall-clock deadline guards `_ws_exec_short` against the same deadlock.

#### Package-install skip for pre-built images (`AGENT_EVAL_K8S_SKIP_PKG_INSTALLS`)
Harbor's agent setup emits `apk add` / `apt-get install` / `yum install` / `npm install -g` / `curl … | bash` commands that require root or internet egress — both unavailable under OpenShift's `restricted-v2` SCC.  
When `AGENT_EVAL_K8S_SKIP_PKG_INSTALLS=1`, `exec()` short-circuits any command matching `_PKG_INSTALL_RE` and returns a no-op result, assuming all dependencies are already baked into the task image.

#### Zombie-process reaping (`tini`)
`sleep infinity` as PID 1 does not reap orphaned child processes created during agent execution.  
The pod command is changed to `tini -- sleep infinity` so `tini` acts as a minimal init and reaps zombies.

#### Auto-load `.env` in `run.py`
`_load_dotenv()` walks from the current working directory up to the filesystem root and loads the first `.env` file found. Uses `os.environ.setdefault` so explicit shell exports always win. This removes the need to `source .env` before running `agent-eval run`.

#### Propagate `workdir` to generated `task.toml`
`tasks.py` now passes `WORKDIR` through to the `task.toml` template, setting `environment.workdir` so Harbor places the agent in the correct working directory inside the trial pod.

## How Has This Been Tested?

- End-to-end trial on OpenShift (RHOAI / Kubernetes with `restricted-v2` SCC): `ng_run → HarborAgent → KubernetesEnvironment → trial pod → claude-code → LLM judge` pipeline completed with reward 0.8125, trajectory of 64 messages returned to NeMo Gym.
- Verified `AGENT_EVAL_K8S_SKIP_PKG_INSTALLS=1` suppresses all root-requiring install commands without affecting agent execution.
- Confirmed HAProxy idle-connection issue no longer kills long-running agent sessions after the fire-and-poll rewrite.

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.


agent_eval/harbor/kubernetes.py
```diff
@@ -301,33 +335,172 @@ async def stop(self, delete: bool) -> None:
...
-        resp = k8s_stream(
-            self._core.connect_get_namespaced_pod_exec,
-            self._pod, self._namespace, container="main",
-            command=["/bin/sh", "-c", command],
-            stderr=True, stdin=False, stdout=True, tty=False,
-            _preload_content=False,
-        )
-        out: list[str] = []
-        err: list[str] = []
+        """Fire-and-poll exec that survives HAProxy idle-connection timeouts.
```
astefanutti: The fire-and-poll rewrite works but adds significant complexity (~200 lines, temp file management, relay processes, daemon threads, and ~1800 exec API calls for a 30-minute agent run). A lighter alternative might be to keep the original single-connection `_ws_exec` and just send WebSocket pings to prevent HAProxy from treating the connection as idle:

```python
def _keepalive(ws, interval=30, stop_event):
    while not stop_event.wait(interval):
        try:
            ws.sock.ping()
        except Exception:
            break
```

HAProxy counts any wire-level frame (including pings) as activity, so a ping every ~30s would keep the connection alive without changing the exec model. That would be ~20 lines instead of ~200, with no temp files, no relay process, and no polling overhead on the API server.

The `_ws_exec_short` daemon-thread-with-hard-deadline pattern is still valuable for the `resp.close()` / `read_channel()` deadlock — but it could be applied directly to the original `_ws_exec` rather than building a full fire-and-poll layer on top.

What do you think?

astefanutti (commented):
@abhijeet-dhumal thanks! I think that's worth trying a websocket ping if it works with HA router.

agent_eval/harbor/kubernetes.py
abhijeet-dhumal: Agreed! I have replaced the fire-and-poll approach with a 30s WebSocket ping as suggested.. Kept the daemon thread wrapper since it solves a different problem (deadlock on resp.close() when HAProxy drops without TCP FIN).

Tested HAProxy router)by running harbor run with KubernetesEnvironment against a real Claude-code agent trial pod  exec completed without WebSocketConnectionClosedException, reward scored. The connection stayed alive well past the 60s HAProxy idle threshold with no polling overhead. ✅

uv.lock
astefanutti, line 1: I think this should be removed.

agent_eval/harbor/run.py
```diff
@@ -27,6 +28,30 @@
 from agent_eval.harbor.reward import _load_score_module
 
 _REPO_ROOT = Path(__file__).resolve().parent.parent.parent
+
+
+def _load_dotenv() -> None:
+    """Load .env from cwd or any ancestor, if present.
```
astefanutti, line 34: Maybe clarify it doesn't handle export KEY=VALUE, inline comments (KEY=val # comment), multiline values, or escaped quotes to avoid someone from assuming full .env spec support.

agent_eval/harbor/kubernetes.py
```diff
@@ -301,44 +335,117 @@ async def stop(self, delete: bool) -> None:
...
+        return result_holder[0] if result_holder else ExecResult(
+            stdout="", stderr="[no result from exec thread]", return_code=1)
 
     async def exec(
         self,
         command: str,
         cwd: str | None = None,
         env: dict[str, str] | None = None,
-        timeout_sec: int | None = None,
+        timeout_sec: int | None = 300,
```
astefanutti: are you sure about this? won't it break long running agents?

abhijeet-dhumal: Would it help to add a shared threading.Lock between the keepalive thread and the resp.update() read loop so ping() and recv() can't race on the socket. The lock is acquired for at most 1s per cycle (the update(timeout=1) window), so keepalive still fires reliably every 30s and there's no contention risk on long-running agents.

abhijeet-dhumal: Done ✅ 
Please review ! 🙌

agent_eval/harbor/kubernetes.py
```diff
@@ -184,7 +218,7 @@ def _pod_manifest(self, image: str, env: dict) -> dict:
         container: dict = {
             "name": "main",
             "image": image,
-            "command": ["sleep", "infinity"],
+            "command": ["tini", "--", "sleep", "infinity"],
```
astefanutti: My understanding is that `tini` was part of the fire-and-poll approach that's been replaced by the keepalive ping. With the current code there are no relay processes to zombie. The remaining exec'd shell commands do produce zombies under `sleep infinity` as PID 1, but:

- The pod is ephemeral — deleted after the trial
- A typical trial produces ~10–50 exec calls, so ~10–50 harmless zombies (no memory/CPU, just a PID slot)
- They all vanish when the pod is deleted

Adding `tini` creates a hard image dependency — if the task image doesn't have it, the pod crashes at startup with a cryptic error. The risk outweighs the benefit for an ephemeral pod. I'd revert to `["sleep", "infinity"]`.

astefanutti (approved):
@abhijeet-dhumal Thanks!

PR #130, opened 2026-06-19
feat(harbor): add grpo_reward judge and 
  configurable score range

Sridhar1030:
## Summary                            
  - Add grpo_reward judge support: when present, its value is used 
  directly as the overall reward                                          
  - Make score range configurable via score_range in eval config instead  
  of hardcoded 1-5                                                        
  - Fall back to legacy gate+average behavior when no grpo_reward judge 
  exists                                                                  
                  
  ## Test plan                                                            
  - Run existing reward tests to verify backward compatibility
  - Test with eval.yaml containing a grpo_reward judge                    
  - Test with custom score_range config

## Release Notes

* **New Features**
  * Added optional `reward` configuration in eval YAML to combine per-judge outputs into a single scalar reward in `[0,1]`.
  * Supports reward modes (`weighted`, single-judge with `score_range`, or expression), optional boolean `gate`, and `raw` judges that bypass normalization.

* **Improvements**
  * Reward resolution now follows a defined precedence order: `reward` config first, then legacy `grpo_reward`, otherwise default gating + averaged normalized numerics.
  * Expression handling is safer: validated at config load, with evaluation failures warning and returning `0.0`.

* **Documentation**
  * Updated configuration documentation and the eval YAML template with the new `reward` section.

* **Tests**
  * Expanded unit/integration coverage for reward parsing, normalization/clamping, gating behavior, formula safety, precedence, and fallback paths.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,6 +109,7 @@ Skills projects create an `eval.yaml` config file with:
 - `traces` — execution data to capture: stdout/stderr, events, metrics (exit code, tokens, cost)
 - `judges` — `builtin` reusable judges (from `agent_eval/judges/`), inline `check` scripts, LLM `prompt`/`prompt_file` (Jinja2 rendered), external `module`/`function`. Optional `arguments` dict for parameterization. Optional `if` condition to skip judges per case based on annotations. Judges receive `outputs["annotations"]` from dataset `annotations.yaml`.
 - `thresholds` — per-judge regression detection. Valid keys: `min_mean`, `min_pass_rate`, `min_win_rate`
+- `reward` — optional. Collapses per-judge results into a single scalar in `[0, 1]` for RL training (GRPO). `formula` selects the mode: `weighted` (weighted sum of `weights`), a single `<judge_name>` (normalized via `score_range`; list in `raw` if already `[0, 1]`), or a Python `<expression>` over judge names (allowed calls: `min`/`max`/`abs`/`round`/`sum`/`len`/`mean`). `gate: true` zeros the reward on any false boolean judge — it gates on *every* boolean judge, so expressions using booleans as their own gate want `gate: false`. Expressions are AST-validated and checked at config load. Resolution order: `reward:` section → legacy `grpo_reward` judge → default (boolean gates + averaged normalized numerics).
 
 Runs are stored in `$AGENT_EVAL_RUNS_DIR` (default `eval/runs`), configured during `/eval-setup`.
 
--- a/agent_eval/config.py
+++ b/agent_eval/config.py
@@ -563,19 +563,41 @@ class EvalConfig:
             )
 
         # Reward composition
-        reward_raw = raw.get("reward")
-        if reward_raw and isinstance(reward_raw, dict):
+        if "reward" in raw:
+            reward_raw = raw.get("reward")
+            if not isinstance(reward_raw, dict):
+                raise ValueError("reward must be a mapping when provided")
             sr = reward_raw.get("score_range", [1, 5])
             if not isinstance(sr, list) or len(sr) != 2:
                 raise ValueError("reward.score_range must be a [min, max] list")
+            try:
+                score_min = float(sr[0])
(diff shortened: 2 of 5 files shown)


astefanutti (commented):
Thanks @Sridhar1030!

Rather than a magic judge name (`grpo_reward`), what about a `reward` section in eval.yaml — same pattern as `thresholds`:

```yaml
reward:
  formula: 0.5 * rfe_quality + 0.3 * feasibility_quality + 0.2 * revision_quality
  gate: true              # boolean judges gate (any fail → 0), default true
  score_range: [1, 5]     # normalization range, default [1, 5]
```

The formula references judge names as variables (normalized to 0-1 via `score_range`). Supports expressions:

```yaml
reward:
  formula: |
    0.6 * rfe_quality + 0.4 * feasibility_quality - 0.1 * (cost / budget)
  gate: true
  score_range: [1, 5]
```

Or for simple cases — `formula: weighted` with explicit weights, or `formula: grpo_reward` to use a single judge directly:

```yaml
reward:
  formula: weighted
  weights:
    rfe_quality: 0.5
    feasibility_quality: 0.3
    revision_quality: 0.2
  gate: true
  score_range: [1, 5]
```

No `reward` section → current default behavior (gate booleans + average numerics). This covers the GRPO use case without tying the composition to a specific judge name, and lets eval authors shape the reward signal for RL training.

Also noted: `config.score_range` in the PR will always be `None` since `EvalConfig` doesn't have that field — a `reward` section would solve that too.

Sridhar1030:
Agreed -- implemented. The `reward:` section now supports three modes:

**Expression** (our GRPO formula):
```yaml
reward:
  formula: |
    gate_ratio = mean([files_exist, frontmatter_valid, ...])
    0.60 * layer1 + 0.30 * layer2 + 0.10 * efficiency
  gate: false
  score_range: [1, 5]
  raw: [efficiency]
```

**Weighted shorthand:**
```yaml
reward:
  formula: weighted
  weights:
    rfe_quality: 0.5
    feasibility_quality: 0.3
  score_range: [1, 5]
```

**Single judge passthrough:**
```yaml
reward:
  formula: rfe_quality
```

Also fixed `config.score_range` being `None` -- now parsed from the `reward:` block. Added `raw:` for judges already in `[0,1]` (like `efficiency`) that shouldn't be normalized. No `reward:` section = existing behavior preserved.

astefanutti (commented):
@Sridhar1030, that looks good overall.

A few findings from Claude Code:

### 1. `exec`/`eval` on formula expressions

`compute_reward_from_config` runs `eval()` (and `exec` for multi-line) on the formula string. The `__builtins__` restriction is good, but `exec` for multi-line opens more surface. Worth documenting that formula expressions are trusted input (from eval.yaml, not user-facing).

### 2. Stale `score_range` dict access in `build_reward`

Lines 252-254 still read `config.score_range` as a dict:
```python
score_range = getattr(config, "score_range", None) or {}
score_min = float(score_range.get("min", _SCORE_MIN_DEFAULT))
```
But `RewardConfig.score_range` is a list `[min, max]`, not a dict. When `reward_cfg` is set, the `score_min`/`score_max` locals are passed to `compose_reward` but unused (the `reward_cfg` path handles its own normalization). When `reward_cfg` is None, `config.score_range` doesn't exist as a top-level attribute. This is effectively dead code — could be cleaned up.

### 3. Weighted formula doesn't normalize by weight sum

If weights are `{a: 0.5, b: 0.3}` (sum 0.8), the result caps at 0.8 even with perfect scores. Either divide by `weight_sum`:
```python
return max(0.0, min(1.0, total / weight_sum)) if weight_sum > 0 else 0.0
```
Or document that weights must sum to 1.0.

### 4. No tests

No tests for `RewardConfig`, `compute_reward_from_config`, or the expression evaluation path. The expression mode especially needs coverage (valid formulas, missing judges, malformed expressions).

astefanutti (approved):
Thanks @Sridhar1030!

PR #133, opened 2026-06-21
feat: add cross-component reference validation to eval-check

Benkapner:
## Summary

Adds reference validation to catch broken cross-component references before they cause runtime failures. Two parts: enhanced `validate_eval.py` checks and a new `reference_checker.py` script for `/eval-check`.

The dependency analysis approach is taken from [harness-eval](https://github.com/redhat-community-ai-tools/harness-eval), a setup evaluation tool for AI code agent configurations.

Builds on: #74 (`/eval-check`)

## Problem

Currently, broken references are only discovered at runtime:
- A judge with `module: my.pkg` + `function: nonexistent` imports fine but crashes when called
- A SKILL.md referencing `${CLAUDE_SKILL_DIR}/scripts/deleted.py` fails mid-execution
- An eval.yaml pointing to a renamed skill name fails at eval startup
- A `check:` judge with invalid Python syntax fails during scoring

These are all catchable statically, before spending time and API credits.

### validate_eval.py: 3 new checks

| Check | What it catches |
|---|---|
| **Judge function existence** | `module` imports OK but `function` doesn't exist or isn't callable |
| **Inline check syntax** | `check:` with invalid Python (ast.parse before runtime) |
| **Plugin dirs** | `runner.plugin_dirs` paths that don't exist or aren't directories |

Note: builtin judge name validation was in the original PR but is now handled by main's `_validate_builtin_arguments()` + `EvalConfig.from_yaml()` schema validation.

### reference_checker.py: cross-component reference validation

New standalone script (no `agent_eval._bootstrap`, matches harness_inventory.py convention):

- **Cross-skill references**: scans SKILL.md files for `/eval-run`, `skills/eval-analyze/` patterns and verifies targets exist. Filters out builtin slash-commands (`/help`, `/clear`, `/config`, etc.) to avoid false positives.
- **Script references**: verifies `${CLAUDE_SKILL_DIR}/scripts/*.py` files exist, including cross-skill paths (`../../scripts/`, `${CLAUDE_PLUGIN_ROOT}/skills/x/scripts/y`)
- **Eval config references**: verifies `execution.skill` (or deprecated top-level `skill:`) resolves to an existing skill
- **Orphan detection**: flags skills not referenced by any component (only in setups with >3 skills)
- Filters out placeholder names (my-skill, skill-name, foo, bar) to avoid false positives from documentation

### eval-check SKILL.md: Step 4f

New "Reference Validation" step that runs `reference_checker.py` and adds findings to the harness health report.

### harness_inventory.py: eval config discovery

New `find_eval_configs()` function discovers eval.yaml files (supporting both `execution.skill` and deprecated top-level `skill:`) and includes them in the inventory count.

## Design decisions

- **Discovery logic is duplicated from eval-check's `harness_inventory.py`**, not imported. There is no cross-skill import convention in this repo, and a shared module would add coupling between two independent skills.
- **No new dependencies.** The scanner uses only stdlib (`re`, `json`, `pathlib`) + optional PyYAML for YAML output (falls back to JSON if PyYAML is unavailable).

## Tested against this repo

```
=== Reference Check ===

References found: 73
Broken references: 1
Missing scripts: 0
Orphan skills: 1
Eval configs: 2
```

73 cross-skill references found (eval-optimize -> eval-run, eval-analyze -> eval-dataset, etc.). One broken reference from the opencode-smoke example eval config pointing to a skill not in this repo. One orphan (eval-check itself, which is user-invoked, not referenced by other skills).

## Test plan

- [x] Run `python3 skills/eval-check/scripts/reference_checker.py --root . --format text` and verify output
- [x] Run with `--format yaml` and verify well-formed YAML
- [x] Verify builtin commands (`/help`, `/clear`) are not flagged as broken refs
- [x] Verify cross-skill script references (`../../scripts/`, `${CLAUDE_PLUGIN_ROOT}/skills/x/scripts/y`) resolve correctly
- [ ] Run on empty directory for graceful handling
- [ ] Run `python3 skills/eval-analyze/scripts/validate_eval.py config eval.yaml` and verify new checks execute
- [ ] Invoke `/eval-check` in Claude Code and verify reference validation appears in the report

## Summary by CodeRabbit

* **New Features**
  * Added reference validation for broken skill and script references, invalid evaluation configuration references, and orphaned skills.
  * Evaluation inventory reports now include discovered configurations and associated skills.
  * Reference analysis supports text, YAML, and JSON report formats.

* **Bug Fixes**
  * Improved validation of inline judges, imported modules, configured functions, and runner plugin directories.
  * Reports now flag malformed configurations and missing or invalid references.

* **Documentation**
  * Updated evaluation-check guidance to describe expanded validation coverage.


skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,378 @@
...
+        if yaml_file.resolve() in seen:
+            continue
+        seen.add(yaml_file.resolve())
+        content = _read_text_safe(yaml_file)
+        if not content:
+            continue
+        try:
+            import yaml
+            parsed = yaml.safe_load(content)
+            if isinstance(parsed, dict) and parsed.get("skill"):
```
astefanutti: **🔴 Blocker — eval-config discovery is a no-op on the current schema.**

`find_eval_configs` gates on `parsed.get("skill")` (top-level only), but `main` moved the skill declaration under `execution.skill` (top-level is deprecated/auto-normalized — see `agent_eval/config.py`, `resolve_skill()` = `execution.skill or top-level skill`). On a current config with only `execution.skill`, this returns `[]`, so `check_eval_config_references` never runs and a broken `execution.skill: nonexistent` reference goes undetected. Verified empirically: `analyze()` returns `eval_configs: []` / `total_references: 0` for such a config.

The PR's "all references resolve / Eval configs: 1" only holds because this branch's own `eval.yaml` still uses the old top-level `skill:`. The branch is behind `main` and needs a rebase.

```python
skill = (parsed.get("execution") or {}).get("skill") or parsed.get("skill")
if isinstance(parsed, dict) and skill:
    configs.append({..., "skill": skill})
```
Same fix is needed in the duplicated copy in `harness_inventory.py` (see other comment).

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,378 @@
...
+import re
+import sys
+from dataclasses import dataclass, field
+from pathlib import Path
+
+MAX_FILE_SIZE = 1_000_000
+
+_SKILL_REF_PATTERNS = [
+    re.compile(r"Skill\s+tool\s+to\s+invoke\s+/(\w[\w-]*)"),
+    re.compile(r"`/(\w[\w-]*)`"),
```
astefanutti: **🔴 Blocker — every backticked slash-command is treated as a skill reference.**

This pattern matches `/help`, `/clear`, `/config`, `/compact`, plugin commands like `/memory-recall`, etc. Since none resolve to a project skill, they land in `broken_refs` and `main()` exits 1. Reproduced: a SKILL.md mentioning `` `/clear` `` and `` `/help` `` produces two "NOT FOUND" broken refs and a non-zero exit. `eval-check` scans arbitrary target projects, where documenting built-in commands in backticks is common — so this false-positive is highly reachable.

Suggest one of: whitelist known built-in/plugin commands, only count a `/name` as broken when it looks like a project skill, or downgrade unknown slash-commands to informational so they don't drive the exit code.

skills/eval-analyze/scripts/validate_eval.py
```diff
@@ -91,9 +112,32 @@ def validate_config(path="eval.yaml"):
...
+            if mod and func_name:
+                fn = getattr(mod, func_name, None)
+                if fn is None:
+                    errors.append(f"judges.{name}.function '{func_name}' not found in module '{module}'")
+                elif not callable(fn):
+                    errors.append(f"judges.{name}.function '{func_name}' in '{module}' is not callable")
+
+        builtin = j.get("builtin", "")
+        if builtin:
+            flat_name = builtin.rsplit("/", 1)[-1] if "/" in builtin else builtin
```
astefanutti: **🟠 Builtin-judge check ignores the category → false negative.**

Stripping the category and matching only the flat name means `builtin: quality/cost_budget` passes validation, but the authoritative `BuiltinJudgeRegistry.get()` (`agent_eval/judges/__init__.py:80-87`) rejects it because `cost_budget` lives in `efficiency/`, not `quality/`. Validate the full `category/name` the way the registry does.

skills/eval-analyze/scripts/validate_eval.py
```diff
@@ -19,6 +20,24 @@
 from find_skills import find_skill
 
 
+def _discover_builtin_judge_names() -> set[str]:
+    """Scan agent_eval/judges/ for available builtin judge names (no module loading)."""
+    judges_dir = Path(__file__).resolve().parents[2] / ".." / "agent_eval" / "judges"
```
astefanutti: **🟠 Re-implements registry discovery instead of importing it.**

This walks the filesystem via `parents[2]/".."/agent_eval/judges` and accepts any `.py`/`.md` stem, whereas the registry requires `.py` judges to define a callable `judge`. The hardcoded relative path also won't resolve under a venv/site-packages/plugin install (where `is_dir()` is False and the check silently passes everything). `validate_eval.py` already activates the venv via `agent_eval._bootstrap`, so just use the source of truth:

```python
from agent_eval.judges import BuiltinJudgeRegistry
r = BuiltinJudgeRegistry(); r.discover()
builtin_names = set(r.list_names())
```

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,378 @@
...
+    re.compile(r"skills/(\w[\w-]*)/SKILL\.md"),
+    re.compile(r"skills/(\w[\w-]*)/"),
+]
+
+_SCRIPT_REF_PATTERN = re.compile(
```
astefanutti: **🟠 Script-ref pattern misses the repo's actual cross-skill references.**

This requires `/scripts/` immediately after `${CLAUDE_SKILL_DIR}`, so it matches none of the ~15 real refs like `${CLAUDE_SKILL_DIR}/../../scripts/discover.py` or `${CLAUDE_SKILL_DIR}/../eval-analyze/scripts/find_skills.py` used across the skills. The check effectively only covers same-skill refs — a broken cross-skill script path is silently missed. Broaden the pattern to allow `../` segments (and resolve the path relative to the skill dir).

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,378 @@
...
+@dataclass
+class ReferenceReport:
+    references: list[Reference] = field(default_factory=list)
+    broken_refs: list[Reference] = field(default_factory=list)
+    missing_scripts: list[Reference] = field(default_factory=list)
+    orphan_skills: list[str] = field(default_factory=list)
+    eval_configs: list[dict] = field(default_factory=list)
+
+
+def _read_text_safe(path: Path) -> str:
```
astefanutti, line 113: **🟡 Duplication that already diverges.**

`_read_text_safe`, `find_skills`, `find_commands`, `find_eval_configs`, and the plugin.json resolution block are copy-pasted from the sibling `harness_inventory.py`, and the copies already differ — e.g. this `_read_text_safe` uses `errors="replace"` while `harness_inventory._read_text_safe` uses bare `read_text()` (different results on non-UTF8 files). Since `harness_inventory.py` is itself standalone (no `agent_eval` import), `reference_checker.py` can import these helpers from it and still honor the "standalone" convention — and the `find_eval_configs` fix then only has to be made once.

skills/eval-check/scripts/harness_inventory.py
```diff
@@ -163,6 +163,34 @@ def find_hooks(root: Path) -> list[dict]:
     return hooks
 
 
+def find_eval_configs(root: Path) -> list[dict]:
```
astefanutti, line 181: **🟡 Second copy of `find_eval_configs` with the same stale-schema bug.**

Identical `parsed.get("skill")` gate as `reference_checker.py`, so the inventory reports `Eval configs: 0` for any current-schema project (verified). Because it's copy-pasted, the fix must be applied in both places or they drift. Prefer a single shared helper that reads `execution.skill`.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,378 @@
...
+    all_refs: list[Reference],
+    configs: list[dict],
+) -> list[str]:
+    if len(skills) <= 3:
+        return []
+
+    referenced_names: set[str] = set()
+    for ref in all_refs:
+        referenced_names.add(ref.target_name)
+        referenced_names.add(ref.source_name)
```
astefanutti: **🟡 Orphan detection under-reports.**

Adding `ref.source_name` to `referenced_names` means a dead entry-point skill that references others but is itself referenced by nobody is *not* flagged — contradicting the "not referenced by any component" label. Reproduced: a source-only skill was excluded while a fully isolated one was flagged. Use only `target_name`, or relabel the finding.

skills/eval-analyze/scripts/validate_eval.py
```diff
@@ -122,6 +166,14 @@ def validate_config(path="eval.yaml"):
         if not sp.exists():
             errors.append(f"runner.settings '{settings}' not found")
 
+    plugin_dirs = runner.get("plugin_dirs") or []
```
astefanutti: **🟡 `plugin_dirs` resolved differently than at runtime.**

Here they're resolved relative to `config_dir`, but the runner resolves them relative to cwd (`agent_eval/agent/claude_code.py:44`, `Path(d).resolve()`). When `--config` points at a subdir, the validator's verdict can disagree with what actually happens at run time.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,378 @@
...
+def _read_text_safe(path: Path) -> str:
+    try:
+        if path.stat().st_size > MAX_FILE_SIZE:
+            return ""
+        return path.read_text(encoding="utf-8", errors="replace")
+    except (OSError, PermissionError, UnicodeDecodeError):
+        return ""
+
+
+def _parse_frontmatter(content: str) -> dict:
```
astefanutti: **🟡 Dead code.** `_parse_frontmatter` is defined but never called (`find_skills` stores raw content instead). It's also a behaviorally-divergent copy of `harness_inventory._parse_frontmatter_description` (strips whitespace around the `---` fences vs exact match). Remove it or wire it up.

astefanutti (commented):
## Automated review — cross-component reference validation

Thanks for this — static reference validation is a genuinely useful addition, and the `validate_eval.py` checks for judge-function existence, inline-check syntax, and `plugin_dirs` are correct and helpful.

Two things worth flagging before merge:

**1. The branch is behind `main` and needs a rebase.** It was cut against the old schema where the skill was a top-level `skill:` key. `main` has since moved it to `execution.skill` (top-level deprecated/auto-normalized), and `validate_eval.py` has been substantially rewritten. This branch's own `eval.yaml` still uses the old form, which is why local testing shows "all resolve / Eval configs: 1".

**2. As a result, the headline eval-config reference check is currently a no-op on current-schema configs** (`find_eval_configs` reads only top-level `skill:`), and the backtick-slash-command pattern produces false-positive broken refs that flip the exit code. Details inline (🔴 = blocker, 🟠 = should fix, 🟡 = minor, 🟢 = nit).

All findings were reproduced against the current-`main` schema before posting. Nice work overall — with the rebase + the two blockers addressed this is a solid check to have.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,378 @@
...
+            continue
+        try:
+            import yaml
+            parsed = yaml.safe_load(content)
+            if isinstance(parsed, dict) and parsed.get("skill"):
+                configs.append({
+                    "path": str(yaml_file.relative_to(root)),
+                    "name": parsed.get("name", yaml_file.parent.name),
+                    "skill": parsed.get("skill", ""),
+                    "judges": parsed.get("judges", []),
```
astefanutti: **🟢 Nit — dead field.** `judges` is parsed and carried here but never read; `analyze()` only uses `skill`/`name`/`path`, and `ReferenceReport.eval_configs` is rebuilt without it. It's also the only field that differs from the `harness_inventory` copy, so the divergence buys nothing.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,378 @@
...
+import sys
+from dataclasses import dataclass, field
+from pathlib import Path
+
+MAX_FILE_SIZE = 1_000_000
+
+_SKILL_REF_PATTERNS = [
+    re.compile(r"Skill\s+tool\s+to\s+invoke\s+/(\w[\w-]*)"),
+    re.compile(r"`/(\w[\w-]*)`"),
+    re.compile(r"skills/(\w[\w-]*)/SKILL\.md"),
```
astefanutti: **🟢 Nit — redundant pattern.** `skills/(\w[\w-]*)/SKILL\.md` is fully subsumed by the next pattern `skills/(\w[\w-]*)/` (same capture group, deduped via the `seen` set), so it can never contribute a reference.

skills/eval-check/SKILL.md
```diff
@@ -91,6 +91,20 @@ Flag any issues found by the inventory script, plus:
...
+Run the reference checker to find broken cross-component references:
+
+```bash
+python3 ${CLAUDE_SKILL_DIR}/scripts/reference_checker.py --root . --format yaml
+```
+
+This checks:
+- **Cross-skill references**: when a SKILL.md mentions `/eval-run` or `skills/eval-analyze`, verify the target skill exists
+- **Script references**: when a SKILL.md references `${CLAUDE_SKILL_DIR}/scripts/foo.py`, verify the script file exists in the skill's directory
+- **Eval config references**: when an eval.yaml names a `skill:`, verify that skill exists
```
astefanutti: **🟢 Doc — reinforces the stale assumption.** "when an eval.yaml names a `skill:`" describes the deprecated top-level form; the canonical location is `execution.skill`. Update alongside the `find_eval_configs` fix so the section actually fires on current configs.

skills/eval-analyze/scripts/validate_eval.py
```diff
@@ -91,9 +112,32 @@ def validate_config(path="eval.yaml"):
...
+            flat_name = builtin.rsplit("/", 1)[-1] if "/" in builtin else builtin
+            if flat_name not in builtin_names:
+                errors.append(
+                    f"judges.{name}.builtin '{builtin}' not found "
+                    f"(available: {', '.join(sorted(builtin_names)) or 'none'})")
+
+        check_expr = j.get("check", "")
+        if check_expr:
+            try:
+                ast.parse(check_expr, mode="exec")
```
astefanutti: **🟢 FYI (benign).** This uses `ast.parse(mode="exec")` on the raw snippet while the runtime wraps it in a function before compiling (`score.py:878`). They happen to agree on the common case — parse-time tolerates the top-level `return ...` form — so genuine syntax errors are still caught. Only an explicitly leading-indented snippet would raise a spurious `IndentationError`. No change required; noting for completeness.

astefanutti (commented):
Thanks @Benkapner, this looks good overall, I've left some comments. Could you please have a look and rebase when you can?

Benkapner (commented):
Thanks @astefanutti . i rebased onto current main and addressed all findings:

Blockers:
- find_eval_configs now checks both execution.skill and deprecated top-level skill:
- Backtick slash-commands filtered via _BUILTIN_COMMANDS allowlist; no more /help//clear false positives

the major stuff:
- Builtin judge name/category validation and registry import: dropped from this PR. Main's validate_eval.py was substantially rewritten and now handles this via _validate_builtin_arguments() + EvalConfig.from_yaml() schema validation, which supersedes the filesystem-walk approach from the original PR
- Script-ref patterns broadened: split into _SCRIPT_REF_SAME_SKILL (handles ../../scripts/) and _SCRIPT_REF_CROSS_SKILL (handles ${CLAUDE_PLUGIN_ROOT}/skills/x/scripts/y), each resolving relative to the correct base

Other fixes:
- Orphan detection: only target_name added to referenced_names
- Dead code removed: _parse_frontmatter, judges field, redundant regex
- SKILL.md updated to reference execution.skill
- validate_eval.py: added function existence check for module judges and plugin_dirs existence check (the two checks from the original PR not already covered by main)

tested against the repo: 73 refs found, 0 script false positives, builtin commands correctly filtered.

Benkapner:
Addressed 4 of 5 CodeRabbit comments in 9f552a3:

1. **CWE-703 (import/getattr exceptions)**: Fixed. Now catches `Exception` (not just `ImportError`) on `import_module`, and wraps `getattr` in try/except for modules with custom `__getattr__`.

2. **CWE-22 (plugin_dirs containment)**: Intentionally not addressed. `plugin_dirs` are developer-controlled paths in their own `eval.yaml` pointing to plugin root directories, which legitimately live outside the project (shared plugins, monorepo layouts). Adding containment would break valid configurations. This is a validation script, not a sandbox.

3. **CWE-22 (eval.yaml symlink containment)**: Fixed. Both `reference_checker.py` and `harness_inventory.py` now skip `eval.yaml` files whose resolved path falls outside the scan root.

4. **Non-string skill type guard**: Fixed. Both files now check `isinstance(skill, str) and skill` before storing, preventing `TypeError` from list/int values in malformed YAML.

5. **Double `scripts/` in output**: Fixed. `target_name` already contains the full relative path; removed the redundant `scripts/` prefix in the format string.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,404 @@
...
+    for cfg in configs:
+        skill_name = cfg["skill"]
+        if skill_name in _PLACEHOLDER_NAMES:
+            continue
+        refs.append(Reference(
+            source_type="eval_config",
+            source_name=cfg["name"],
+            target_type="skill",
+            target_name=skill_name,
+            exists=skill_name in known_skill_names,
```
astefanutti: **🔴 Blocker — the checker now fails on this repo (exit 1).** Running it on the PR head:

```
References found: 73
Broken references: 1
  eval_config/opencode-smoke -> skill/opencode-smoke (NOT FOUND)
Orphan skills: 1  (eval-check)
$ echo $?   # 1
```

This is a false positive: `eval/opencode-smoke/eval.yaml` sets `runner.type: cli`, where `execution.skill` is just the `{agent}` label substituted into a bash command template (`cli_runner.py:103`) — no Claude Code SKILL.md is involved. Fixing blocker #1 correctly surfaced this config, but `check_eval_config_references` doesn't consider `runner.type`.

Suggest skipping config skill-refs when `runner.type` isn't `claude-code` (or downgrading them to informational).

skills/eval-analyze/scripts/validate_eval.py
```diff
@@ -637,6 +655,14 @@ def validate_config(path="eval.yaml"):
         if not sp.exists():
             errors.append(f"runner.settings '{settings}' not found")
 
+    plugin_dirs = runner.get("plugin_dirs") or []
```
astefanutti: **🔴 Blocker — unhandled `TypeError` crashes the whole validator.** No type guard on the entries:

```yaml
runner:
  plugin_dirs:
    -            # dangling item, e.g. after commenting out the only path
```
```
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

A scalar (non-list) value is also iterated character-by-character. Reproduced against this head. Please guard:

```python
plugin_dirs = runner.get("plugin_dirs") or []
if isinstance(plugin_dirs, str):
    plugin_dirs = [plugin_dirs]
for i, pd in enumerate(plugin_dirs):
    if not isinstance(pd, str) or not pd:
        errors.append(f"runner.plugin_dirs[{i}] must be a non-empty string")
        continue
```

astefanutti (commented):
## Re-review — rebased head (9f552a3)

Thanks for the thorough turnaround. I re-verified all 14 prior findings against the new head and re-ran the checker; **9 are fully fixed** (script-ref patterns, orphan `target_name`, `harness_inventory` schema, dead `_parse_frontmatter`/`judges`/regex, SKILL.md wording, and the `_discover_builtin_judge_names` walk removed by dropping that check). The rebase onto main is clean. 

Two things block merge, and both are new rather than leftovers:

**1. The checker now exits 1 on this repo.** Fixing the schema read correctly surfaced `eval/opencode-smoke/eval.yaml`, but that config uses `runner.type: cli`, where `execution.skill` is just a label for a bash command template — so it's reported as a broken reference. A health-check that fails on its own repo will be disabled fast.

**2. New unhandled `TypeError`** in the `plugin_dirs` loop crashes the entire validator on a malformed (`None`/scalar) entry.

Beyond those, the reference resolver has a few systematic false-positive/negative classes worth closing before this becomes an enforced gate: slash commands aren't valid targets, plugin-namespaced names (`rfe.create` — the repo's own template example) always fail, the built-in allowlist both misses ~22 real built-ins and suppresses real project skills named `review`/`init`, and cross-skill script refs always resolve against `root/skills` (breaking `.claude/skills/` layouts and matching GitHub URLs). Details inline.

**On dropping the builtin-judge name/category check** — your call is reasonable, and I'd accept it, but the supersession is only partial, so it's worth recording as a follow-up rather than "covered":
- `builtin: typo_name` → main emits a **WARNING** and **exits 0** (the original PR made it a hard error). `EvalConfig.from_yaml()` stores `builtin` as a plain string and never consults the registry.
- The category form is still unhandled on both sides: `quality/cost_budget` (invalid — it's in `efficiency/`) and `efficiency/cost_budget` (valid) produce the *same* "unknown builtin" warning, because `BUILTIN_ARGS` is keyed by flat name. So the valid FQN form gets a spurious warning and silently skips argument validation.

That last one is a pre-existing main bug, not yours — happy for it to be a separate issue.

All findings above were reproduced against this head before posting. 🔴 = blocker, 🟠 = should fix, 🟡 = minor, 🟢 = nit.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,404 @@
...
+                found_names.add(name)
+
+        for ref_name in found_names:
+            if ref_name == comp["name"] or ref_name in _PLACEHOLDER_NAMES:
+                continue
+            key = (comp_type, comp["name"], "skill", ref_name)
+            if key in seen:
+                continue
+            seen.add(key)
+            exists = ref_name in known_skill_names
```
astefanutti: **🟠 Slash commands aren't valid reference targets.** `known_skill_names` holds only skills, so a `` `/deploy` `` reference backed by a real `commands/deploy.md` is reported broken — even though `find_commands()` already discovered it (it's used as a reference *source* but never as a target). Reproduced: fixture with `commands/deploy.md` + a SKILL.md mentioning `` `/deploy` `` → `skill/a -> skill/deploy (NOT FOUND)`.

Fix: resolve targets against skills ∪ commands.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,404 @@
...
+_SCRIPT_REF_CROSS_SKILL = re.compile(
+    r"(?:\$\{CLAUDE_PLUGIN_ROOT\}/)?skills/([\w-]+/scripts/\S+\.py)"
+)
+
+_PLACEHOLDER_NAMES = {
+    "skill-name", "my-skill", "my-skill-name", "name", "foo", "bar",
+    "example", "your-skill", "target-skill",
+}
+
+_BUILTIN_COMMANDS = {
```
astefanutti: **🟠 The built-in allowlist cuts both ways.**

*False positives* — ~22 real built-ins are missing, each still reported broken when documented in backticks: `/model`, `/agents`, `/mcp`, `/context`, `/hooks`, `/add-dir`, `/resume`, `/export`, `/usage`, `/rewind`, `/todos`, `/statusline`, `/output-style`, `/pr-comments`, `/release-notes`, `/plugin`, `/sandbox`, `/effort`, `/exit`, `/privacy-settings`, `/upgrade`, `/ide`.

*False negatives* — the list also suppresses names that are commonly **real project skills**: `review`, `init`, `memory`, `config`, `status`, `bug`. Reproduced: a project with a genuine `skills/review/` referenced from another SKILL.md has that reference silently dropped, and `review` then shows up as an "orphan".

Since a hand-maintained list can't be complete, consider resolving against discovered skills+commands first and only treating *unresolved* names as informational (rather than broken), instead of hardcoding.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,404 @@
...
+        skill_dir = skill["dir"]
+        content = skill["content"]
+
+        matches: list[tuple[str, Path]] = []
+        for match in _SCRIPT_REF_SAME_SKILL.finditer(content):
+            rel = match.group(1)
+            matches.append((rel, (skill_dir / rel).resolve()))
+        for match in _SCRIPT_REF_CROSS_SKILL.finditer(content):
+            rel = match.group(1)
+            matches.append((f"skills/{rel}", (root / "skills" / rel).resolve()))
```
astefanutti: **🟠 Cross-skill script refs always resolve against `root/skills`.** Because the `${CLAUDE_PLUGIN_ROOT}/` prefix is optional in the pattern, any incidental `skills/<x>/scripts/<y>.py` substring is captured and resolved against `root/skills` — which is wrong for the `.claude/skills/` layout that `find_skills()` explicitly supports, and for URLs.

Reproduced (a) `.claude/skills/` project: a SKILL.md referencing its own existing `.claude/skills/alpha/scripts/run.py` → `Missing scripts: 1` (false positive, exit 1). (b) A GitHub URL containing `.../skills/analyzer/scripts/deps.py` → both a false missing-script *and* a false broken skill ref.

Resolve relative to the actual skills base that `find_skills()` used, and require the `${CLAUDE_PLUGIN_ROOT}/` prefix (or anchor the match) so URLs/prose don't match.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,404 @@
...
+            ))
+    return refs
+
+
+def check_eval_config_references(
+    configs: list[dict],
+    known_skill_names: set[str],
+) -> list[Reference]:
+    refs: list[Reference] = []
+    for cfg in configs:
```
astefanutti, line 365: **🟠 Plugin-namespaced skill names are always reported broken.** Skill names are matched literally against local directory basenames, so the `plugin.skill` form fails — including this repo's own canonical example, `skill: rfe.create` (`skills/eval-analyze/references/eval-yaml-template.md:46`), and any skill installed from an external plugin dir/marketplace rather than living in-tree.

Reproduced: `execution.skill: rfe.create` with `skills/create/SKILL.md` present → `eval_config/e -> skill/rfe.create (NOT FOUND)`.

Suggest matching on the post-`.` segment (or skipping namespaced/external names rather than calling them broken).

Also here: no dedup (unlike `check_skill_references`' `seen` set), so generated Harbor task packages — one `tests/eval.yaml` per case, each keeping `execution` verbatim — multiply one reference into N identical entries. And `execution.steps[].skill` (multi-step configs, `config.py:285`) is never read, so those refs go unvalidated.

skills/eval-analyze/scripts/validate_eval.py
```diff
@@ -572,11 +576,25 @@ def validate_config(path="eval.yaml"):
                 warnings.append(f"judges.{name}.context '{ctx_file}' not found")
         module = j.get("module", "")
         if module:
+            mod = None
             try:
                 import importlib
-                importlib.import_module(module)
+                mod = importlib.import_module(module)
             except ImportError:
                 errors.append(f"judges.{name}.module '{module}' not importable")
```
astefanutti, line 591: **🟠 The new function-existence check never fires for project-local judge modules.** `validate_eval.py` doesn't put the project root on `sys.path`, but the runtime does (`score.py:1251-1252`). So a perfectly valid project-local judge module reports a hard **ERROR** and the config is declared INVALID:

```
ERROR: judges.local_mod.module 'myjudges.custom' not importable
INVALID: t
```
(reproduced with `myjudges/custom.py` at the project root — imports fine at run time)

The `not importable` error itself is pre-existing on main, but it means the *new* `function:` check is dead exactly where it's most useful: `mod` is `None`, so the getattr check is skipped. Inserting the config/project root into `sys.path` before importing would make both work.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,404 @@
...
+    skills: list[dict],
+    all_refs: list[Reference],
+    configs: list[dict],
+) -> list[str]:
+    if len(skills) <= 3:
+        return []
+
+    referenced_names: set[str] = set()
+    for ref in all_refs:
+        referenced_names.add(ref.target_name)
```
astefanutti, line 399: **🟡 Orphan detection is now noisy for entry-point skills.** With only `target_name` counted (the change I suggested — thanks), liveness means "referenced by another skill/command". But README.md and CLAUDE.md aren't scanned as reference sources, so user-facing entry points get flagged: on this very repo, `eval-check` is reported as an orphan despite being documented in README/CLAUDE.md and marked `user-invocable: true` in its own frontmatter.

Suggest also scanning README/CLAUDE.md as sources, or exempting `user-invocable` skills. (Orphans don't affect the exit code, so this is report noise rather than a failure.)

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,404 @@
...
+
+        for display_path, resolved in matches:
+            base_name = resolved.name
+            if base_name.split(".")[0] in _PLACEHOLDER_NAMES:
+                continue
+            key = (skill["name"], display_path)
+            if key in seen:
+                continue
+            seen.add(key)
+            if not resolved.is_relative_to(root_resolved):
```
astefanutti, line 336: **🟡 Out-of-root script refs are silently dropped.** A ref that escapes the root is `continue`d — neither reported nor counted. Reproduced: `${CLAUDE_SKILL_DIR}/../../../shared/scripts/typo_tool.py` (exists nowhere) → `total_references: 1`, `missing_scripts: []`, exit 0. A genuinely broken path becomes invisible; better to surface it as out-of-root/unverified.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,404 @@
...
+    re.compile(r"Skill\s+tool\s+to\s+invoke\s+/(\w[\w-]*)"),
+    re.compile(r"skills/(\w[\w-]*)/"),
+]
+
+_BACKTICK_SLASH_PATTERN = re.compile(r"`/(\w[\w-]*)`")
+
+_SCRIPT_REF_SAME_SKILL = re.compile(
+    r"\$\{CLAUDE_SKILL_DIR\}/((?:\.\./)*(?:[\w-]+/)*scripts/\S+\.py)"
+)
+_SCRIPT_REF_CROSS_SKILL = re.compile(
```
astefanutti, line 45: **🟢 Nit — greedy `\S+\.py` merges adjacent refs.** Two script refs on one line with no intervening whitespace collapse into one bogus path:

`${CLAUDE_SKILL_DIR}/scripts/a.py,${CLAUDE_SKILL_DIR}/scripts/b.py` → captures `scripts/a.py,${CLAUDE_SKILL_DIR}/scripts/b.py`, reported as a missing script. `[\w./-]+\.py` would be tighter.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,404 @@
...
+            if content:
+                commands.append({
+                    "name": name,
+                    "path": str(md_file.relative_to(root)),
+                    "content": content,
+                })
+    return commands
+
+
+def find_eval_configs(root: Path) -> list[dict]:
```
astefanutti, line 202: **🟢 Duplication is still present — and grew.** `find_eval_configs` is now a ~35-line near-verbatim third copy of the same logic (here vs `harness_inventory.py:166`); the schema fix had to be applied twice, which is exactly how the original divergence arose. `_read_text_safe`, `find_skills`, `find_commands`, and the plugin.json block remain duplicated and still differ behaviourally — this copy uses `errors="replace"` while `harness_inventory.py` uses bare `read_text()`, so a single non-UTF8 byte in a SKILL.md makes one script see the skill and the other drop it entirely.

Not a blocker, but importing the shared helpers from `harness_inventory` would prevent the next drift.

skills/eval-check/SKILL.md
```diff
@@ -91,6 +91,20 @@ Flag any issues found by the inventory script, plus:
 - Skills with very broad trigger descriptions that overlap with many peers
 - Custom commands that shadow built-in Claude Code commands
 
+### 4f. Reference Validation
```
astefanutti, line 95: **🟢 Docs — three small mismatches in Step 4f.**
1. It doesn't mention that the script **exits 1** whenever it finds anything, unlike the sibling `harness_inventory.py` which always exits 0. That conflicts with eval-check's "informational report" contract and will trip any CI wrapper.
2. "when a SKILL.md mentions `/eval-run`" overstates the matching — only backticked `` `/name` ``, `skills/name/`, or the literal phrase "Skill tool to invoke /name" match. A plain-text `Use /ghost-skill` is invisible (verified: 0 references found).
3. `website/guides/eval-check.md:89` still says "Findings are grouped into five categories" and omits Reference Validation — worth updating alongside this.

skills/eval-check/scripts/harness_inventory.py
```diff
@@ -163,6 +163,43 @@ def find_hooks(root: Path) -> list[dict]:
...
```
astefanutti, line 181: **🟢 Nit — "Eval configs" undercounts.** Only configs declaring a skill are collected, so prompt-mode and cli-runner configs are omitted: this repo has 4 `eval.yaml` files but the inventory prints `Eval configs: 2`. The new field also isn't surfaced anywhere in the report template.

astefanutti:
@Benkapner thanks, I've left a couple of other comments. Please have a look when you can and it'll be good to go.

Benkapner:
@astefanutti  All findings addressed in 2399531:

Blockers:
- Eval configs with runner.type != "claude-code" are now skipped (opencode-smoke was a false positive; execution.skill there is just a label for a bash command template)
- plugin_dirs entries are now type-guarded: handles None, scalar strings, and non-string values without crashing

the Should-fix:
 Dropped the hardcoded _BUILTIN_COMMANDS allowlist entirely. Backtick /name targets now resolve against discovered skills AND commands. Unresolved names are reported as broken (no silent suppression of real project skills like review, init, memory)
- Plugin-namespaced skill names (e.g., rfe.create) matched by their local segment (create)
- Cross-skill script pattern now requires the ${CLAUDE_PLUGIN_ROOT}/ prefix (no longer optional), preventing false matches on .claude/skills/ layouts and URLs
- Project root added to sys.path before importing judge modules, matching the runtime behavior in score.py:1251-1252
- user-invocable: true skills exempt from orphan detection

Minor:
- Out-of-root script refs surfaced as broken (not silently dropped)
- Script ref regex tightened from \S+ to [\w./-]+ to prevent greedy merging

Docs:
- SKILL.md Step 4f updated with exit code, matching behavior, and runner-type filtering
- Website guide updated: "six categories" (not five), with Reference Validation row

On the builtin-judge category/name issue: agreed that's a pre-existing main bug, not this PR's scope. Happy for it to be a separate issue.

On duplication: acknowledged. Both scripts are in the same skill directory so the coupling is contained, but the behavioral divergence (errors="replace" vs bare read_text()) is a fair point. Importing shared helpers within the same skill would be cleaner; deferring to a follow-up to keep this PR focused on the review fixes.

tested: 72 refs found, 0 broken, 0 orphans, 0 missing scripts, exit 0.

Benkapner:
Addressed all 4 CodeRabbit comments in 131decf:

1. check_code guarded as string before ast.parse (non-string values like check: 42 now report a type error instead of crashing)
2. Judge module imports now use Path.cwd() instead of config_dir, matching the runtime behavior in score.py:1251-1252
3. plugin_dirs container validated before iteration: rejects int, dict, None with a clear error; accepts list or scalar string
4. Added ### Reference Validation to the website guide report example (was claiming six categories but only showing five headings)

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,446 @@
...
+from pathlib import Path
+
+MAX_FILE_SIZE = 1_000_000
+
+_SKILL_REF_PATTERNS = [
+    re.compile(r"Skill\s+tool\s+to\s+invoke\s+/(\w[\w-]*)"),
+    re.compile(r"skills/(\w[\w-]*)/"),
+]
+
+_BACKTICK_SLASH_PATTERN = re.compile(r"`/(\w[\w-]*)`")
```
astefanutti: **🔴 Blocker — dropping the allowlist reintroduces the round-1 false positive, and it's severe.** Every backticked `/name` now resolves against local skills+commands only, with no notion of agent built-ins, so any SKILL.md documenting `/help`, `/model`, `/clear`, `/mcp`, `/hooks`, `/cost`, `/config` is reported broken and exits 1.

Measured against **Anthropic's own official `plugin-dev` plugin**:
```
References found: 9
Broken references: 9     <-- 100% false-positive rate
  skill/command-development -> skill/help      (NOT FOUND)
  skill/mcp-integration     -> skill/mcp       (NOT FOUND)
  skill/hook-development    -> skill/hooks     (NOT FOUND)
  ...
```
This repo passes only because its own SKILL.mds happen not to backtick a built-in.

My round-2 suggestion was two halves and only the first landed: resolve against skills ∪ commands (done ✅), **and** treat *unresolved* refs from the low-confidence backtick pattern as informational rather than broken. A bare `` `/name` `` in prose is indistinguishable from a built-in, another plugin's command, or a filesystem path. Suggest: keep hard-broken for the high-confidence patterns (`skills/x/`, "Skill tool to invoke /x"), and report unresolved backtick-only names in a separate non-failing "unresolved" bucket.

astefanutti (commented):
## Round-3 re-review — head 131decf

Good progress: **6 of 13** round-2 findings are fully fixed and I verified each. The `plugin_dirs` crash is gone (tested `null`/`42`/`{}`/`[123]`/`[null]` — clean errors, no traceback), the `runner.type` filter works (including the default-to-claude-code case), commands are now valid reference targets, the greedy-regex merge is fixed, the `user-invocable` exemption works, and the checker is green on this repo (72 refs, 0 broken, exit 0).

I also agree with your calls on the builtin-judge category issue (separate issue) and the duplication deferral. And you were right about `Path.cwd()` for judge modules — I checked `score.py:2184` and it uses cwd too, so that's correct parity, not a bug. I withdraw that one.

**However, two of the fixes introduced worse regressions than the problems they solved**, and I only caught them by running the checker against real projects outside this repo:

**1. Dropping `_BUILTIN_COMMANDS` → 100% false-positive rate on Anthropic's own `plugin-dev` plugin** (9 references found, 9 reported broken). My round-2 comment had two halves and only the first landed — resolving against skills ∪ commands was right, but unresolved *backtick-only* refs need to be informational, not broken, because a bare `` `/name` `` in prose is indistinguishable from a built-in or another plugin's command.

**2. `rsplit(".")` without a fallback → 21 false broken refs on rfe-creator**, whose skill dirs are literally named `rfe.create` / `rfe.speedrun` / etc. That one's on me — my suggestion should have said *fallback*, not *replace*. One-line fix inline.

Same run surfaced that generated Harbor task packages are swept up (`Eval configs: 21` for one real config), and that the `.claude/skills` fallback in the cross-skill base loop is unreachable dead code.

Suggestion for confidence before merge: run `reference_checker.py --root <path>` against two or three *external* projects (rfe-creator, an installed plugin) and require exit 0. Passing on this repo alone hasn't been a reliable signal in any round so far.

All findings reproduced against this head. 🔴 blocker, 🟠 should fix, 🟡 minor, 🟢 nit.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,446 @@
...
+        skill_name = cfg["skill"]
+        if skill_name in _PLACEHOLDER_NAMES:
+            continue
+        key = (cfg["path"], skill_name)
+        if key in seen:
+            continue
+        seen.add(key)
+        if cfg["runner_type"] != "claude-code":
+            continue
+        local_name = skill_name.rsplit(".", 1)[-1] if "." in skill_name else skill_name
```
astefanutti: **🔴 Blocker — `rsplit(".")` has no fallback, breaking the real plugin-skill convention.** `local_name` replaces the full name instead of being tried as an alternative, so skill directories that *literally* contain a dot are now reported broken even though `find_skills()` discovered them.

That's the actual convention in the sibling **rfe-creator** project (`rfe.create`, `rfe.speedrun`, `rfe.review`, `rfe.split`, `rfe.submit`, `rfe.auto-fix`, `rfe-creator.update-deps`):
```
$ reference_checker.py --root ~/Development/opendatahub-io/rfe-creator
Broken references: 21
  eval_config/rfe-speedrun -> skill/rfe.speedrun (NOT FOUND)   # the dir IS named rfe.speedrun
exit 1
```
This is a regression from my own round-2 suggestion — sorry, I should have been explicit that it needed to be a *fallback*:
```python
exists = skill_name in known_names or skill_name.rsplit(".", 1)[-1] in known_names
```
Also worth handling the `plugin:skill` colon form (documented in this repo's CLAUDE.md) the same way — it isn't split at all today, so a valid `diagram-skills:skill-diagram` reference reports broken.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,446 @@
...
+                    "content": content,
+                })
+    return commands
+
+
+def find_eval_configs(root: Path) -> list[dict]:
+    configs = []
+    seen: set[Path] = set()
+    root_resolved = root.resolve()
+    for yaml_file in root.rglob("eval.yaml"):
```
astefanutti, line 206: **🟠 Generated Harbor task packages are swept up and duplicated.** `rglob("eval.yaml")` matches every `.../steps/*/tests/eval.yaml` written by `agent_eval/harbor/tasks.py`, and the dedup key includes the path, so one logical config is counted N times. On rfe-creator: **`Eval configs: 21`** — one real `eval.yaml` plus 20 copies under `eval/harbor-tasks/case-XXX/steps/submit/tests/eval.yaml`, each emitting an identical broken reference.

Suggest skipping generated task-package dirs (`harbor-tasks/`, `*/tests/eval.yaml` inside them) and deduping config refs by `(skill, runner_type)` rather than path.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,446 @@
...
+        skill_dir = skill["dir"]
+        content = skill["content"]
+
+        matches: list[tuple[str, Path]] = []
+        for match in _SCRIPT_REF_SAME_SKILL.finditer(content):
+            rel = match.group(1)
+            matches.append((rel, (skill_dir / rel).resolve()))
+        for match in _SCRIPT_REF_CROSS_SKILL.finditer(content):
+            rel = match.group(1)
+            for base in skills_bases:
```
astefanutti: **🟠 The `.claude/skills` fallback is unreachable.** The loop breaks on the first candidate that is `is_relative_to(root)` — but `root/"skills"/rel` is *always* in-root whether or not it exists, so `skills_bases[1]` is dead code and `.claude/skills/`-layout projects report valid scripts as missing.

Reproduced: `.claude/skills/alpha/SKILL.md` referencing `${CLAUDE_PLUGIN_ROOT}/skills/beta/scripts/helper.py`, with the file present at `.claude/skills/beta/scripts/helper.py` → `Missing scripts: 1`, exit 1. Same for plugin.json-declared skill roots.

Fix: prefer `candidate.exists()` when choosing the base, falling back to the first in-root candidate only for reporting.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,446 @@
...
+            key = (skill["name"], display_path)
+            if key in seen:
+                continue
+            seen.add(key)
+            if not resolved.is_relative_to(root_resolved):
+                refs.append(Reference(
+                    source_type="skill",
+                    source_name=skill["name"],
+                    target_type="script",
+                    target_name=f"{display_path} (outside project root)",
```
astefanutti, line 344: **🟠 The out-of-root fix only covers one branch.** `${CLAUDE_SKILL_DIR}` refs now surface as "(outside project root)" ✅, but a `${CLAUDE_PLUGIN_ROOT}/skills/...` ref that escapes root under *both* bases never reaches this branch — the base loop at line 252 appends nothing, so it's still silently dropped.

Reproduced: `${CLAUDE_PLUGIN_ROOT}/skills/beta/scripts/../../../../../outside/evil.py` → `missing_scripts: []`, no diagnostic.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,446 @@
...
+        if not content:
+            continue
+        try:
+            import yaml
+            parsed = yaml.safe_load(content)
+            if not isinstance(parsed, dict):
+                continue
+            execution = parsed.get("execution") or {}
+            runner = parsed.get("runner") or {}
+            runner_type = runner.get("type", "claude-code")
```
astefanutti, line 227: **🟡 A non-mapping `runner:` silently drops the whole config.** This call sits inside the broad `except Exception: continue`, so `runner: claude-code` (a scalar) raises `AttributeError` and the config vanishes — not listed, not checked, no diagnostic. A genuinely broken skill reference in that file is invisible. Worth guarding with `isinstance(runner, dict)` and reporting unparseable configs rather than swallowing them.

Same block: `execution.steps[].skill` (multi-step configs, `config.py:285`) is still never read, so those references go unvalidated — a steps-only config is dropped entirely by the `isinstance(skill, str)` guard.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,446 @@
...
+            source_type="eval_config",
+            source_name=cfg["name"],
+            target_type="skill",
+            target_name=skill_name,
+            exists=local_name in known_skill_names,
+        ))
+    return refs
+
+
+def find_orphan_skills(
```
astefanutti, line 389: **🟡 Orphan detection has two remaining inconsistencies.**
1. It only credits eval configs whose `runner_type == "claude-code"`, so a skill referenced *solely* by a `cli`/`opencode` config is listed as an orphan in the same output that lists the config using it.
2. The `user-invocable: true` exemption is effectively blanket — 9 of this repo's 10 skills declare it, so the check can only ever fire for `eval-anova`. The reported "0 orphans" reflects the exemption, not a clean graph.

Also, only SKILL.md and `commands/*.md` count as sources, so skills referenced from prompt files or `references/` are false orphans — rfe-creator reports 5 orphans including `assess-rfe`, which *is* referenced from `rfe.review/SKILL.md`'s prompt assets.

skills/eval-check/scripts/reference_checker.py
```diff
@@ -0,0 +1,446 @@
...
+        if not resolved.is_relative_to(root_resolved):
+            continue
+        if resolved in seen:
+            continue
+        seen.add(resolved)
+        content = _read_text_safe(yaml_file)
+        if not content:
+            continue
+        try:
+            import yaml
```
astefanutti, line 221: **🟡 Missing PyYAML silently disables two checks.** This script has no `agent_eval._bootstrap` venv activation, and the website guide documents PyYAML as optional. Without it, `find_eval_configs` and `_parse_frontmatter` swallow the `ImportError` and return empty — so eval-config validation and the `user-invocable` orphan exemption quietly turn off while the report still says "All references resolve." Better to warn once.

skills/eval-analyze/scripts/validate_eval.py
```diff
@@ -637,6 +662,24 @@ def validate_config(path="eval.yaml"):
         if not sp.exists():
             errors.append(f"runner.settings '{settings}' not found")
 
+    plugin_dirs = runner.get("plugin_dirs")
+    if plugin_dirs is None:
+        plugin_dirs = []
+    elif isinstance(plugin_dirs, str):
```
astefanutti, line 668: **🟢 Nit — the scalar normalization diverges from runtime.** Coercing a scalar `plugin_dirs: plugins` to `["plugins"]` makes the validator pass clean, but `agent_eval/config.py` iterates the string character-by-character at runtime. Validating a shape the runtime mishandles is the one case worth rejecting outright rather than normalizing.

(The rest of the guard is solid — I tested `null`, `42`, `{a: 1}`, `[123]`, `[null]`: all produce clear errors, no traceback. The round-2 crash is properly fixed.)

skills/eval-check/SKILL.md
astefanutti, line 95: **🟢 Docs — Step 1's inventory list wasn't updated.** `harness_inventory.py` now emits `Eval configs: N` (text) and an `eval_configs` key (yaml), but Step 1's "This reports:" list and the report template's Inventory section don't mention it, so the new field never reaches the report.

website/guides/eval-check.md
```diff
@@ -124,6 +125,7 @@ Generated: <date>
 ### CLAUDE.md Duplication
 ### Type Misclassification
 ### Structural Issues
+### Reference Validation
```
astefanutti, line 141: **🟢 Docs — the guide never documents `reference_checker.py` itself.** The category table gained a row, but unlike Step 1 (which shows the `harness_inventory.py` command), the guide doesn't show this script's invocation or its non-zero exit contract, and the mermaid flow still names only `harness_inventory.py`.

astefanutti:
@Benkapner I pushed d8adc53 to this branch directly (maintainer push) rather than sending you round-4 comments — it fixes the two blockers plus the three should-fix items from my last review. Please review it as you would any patch; happy to drop it if you'd rather take a different approach.

**What changed**

- **Backticked `/name` is now low-confidence.** An unresolved one is reported under a new *unresolved references* bucket (informational, exit 0) instead of broken. Explicit `skills/<name>/` paths and the "Skill tool to invoke /name" phrasing remain hard errors, so real breakage is still caught. This keeps the benefit of dropping the allowlist — a real project skill named `review`/`init` still resolves — without the false positives.
- **Namespaced names resolve both ways** (`_resolves()`): full name *or* local segment, and the `plugin:skill` form. The previous `rsplit` replaced rather than fell back, which broke directories literally named `rfe.create`.
- **Generated Harbor task packages are skipped**, detected structurally (an `eval.yaml` under `tests/` with a `task.toml` in an ancestor) rather than by directory name, since `--out` is configurable. Config refs now dedup on the reference, not the file. Applied to `harness_inventory.py` too.
- **Cross-skill script base picks where the script exists**, making the `.claude/skills` base reachable.
- **`${CLAUDE_PLUGIN_ROOT}` refs escaping the root** are reported as unverifiable instead of silently dropped.

**Verification** — the three projects that previously failed now all exit 0:

| Project | Before | After |
|---|---|---|
| this repo | 0 broken, exit 0 | 0 broken, 0 unresolved, exit 0 |
| rfe-creator | **21 broken**, 21 eval configs, exit 1 | 0 broken, 1 unresolved, **1** eval config, exit 0 |
| Anthropic `plugin-dev` | **9/9 broken**, exit 1 | 0 broken, 10 unresolved, exit 0 |

Plus a 13-case regression suite asserting true positives still fire (broken `skills/ghost/` ref, missing script, broken eval-config skill, namespaced ref to a genuinely missing skill) — all pass, and `pytest tests/` is 853 passed / 2 skipped.

Two notes for you to decide on:
1. I did **not** touch the remaining minor items from my last review (non-mapping `runner:` swallowed, `execution.steps[].skill` unread, cli-config-referenced skills counted as orphans, missing-PyYAML silently disabling checks, `plugin_dirs` scalar normalization, and the Step 1 inventory doc line). Happy for those to be follow-ups.
2. Still worth filing the two agreed follow-ups: the builtin-judge name/category gap (pre-existing main bug) and the `reference_checker.py` ↔ `harness_inventory.py` duplication.

From my side this is good to merge once you're happy with the patch.

Benkapner:
Thanks @astefanutti, happy with the changes.

astefanutti (approved):
Thanks @Benkapner, that's great!

PR #146, opened 2026-07-03
feat(report): add per-case reward overview table and Harbor scoring improvements

Sridhar1030:
## Summary

Adds a **Per-Case Reward Overview** to the eval report and a set of Harbor-mode scoring improvements. Since the initial version, four follow-up commits (`f1ed86a`, `e97cb5e`, `42666e8`, `1813991`, `4a9cfe1`) hardened correctness, made the code runner-agnostic, and eliminated redundant I/O.

### `report.py` — Per-case reward overview

- New "Per-Case Reward Overview" HTML section: reward + all judge scores per case, with color-coded Gate / LLM / Other columns and an average row that reports `N of M scored` when some cases are missing rewards.
- Judge classification reuses the authoritative `judge_type` stamped per-case by `score.py` — `check` → Gate, `llm`/`builtin`-LLM → LLM, everything else → Other. No config-`type`-field guessing.
- Builtin judges are further disambiguated via `agent_eval.judges.BuiltinJudgeRegistry.kind`: prompt-based builtins (`quality/`, `safety/`) group with LLM (1-5 default), Python-callable builtins (`efficiency/`, `process/`) with Other (0-1 default). No hardcoded `efficiency` special-case.
- Color bands are driven by each judge's configured `score_range` (fraction of the range → pass/warn/fail), with sensible defaults per group.
- Reward computation delegates to `agent_eval.harbor.reward.compose_reward` and now threads `RewardConfig` from `EvalConfig.from_yaml(...)` via a new `report.load_reward_cfg()` helper — the report's reward column matches what `build_reward` trains on for `reward: { judge / weighted / formula / gate }` configs. Fully-skipped cases render `—` instead of a misleading `1.0`.
- Defensive HTML escaping; `_reward_cell` handles `None` and non-finite floats gracefully.

### `score.py` — Harbor scoring compatibility

- `load_case_record` supports both `events.json` (flat) and `events.jsonl` (raw Claude Code session transcripts from Harbor pods). `.jsonl` inputs are normalized through the canonical `agent_eval.events.parse_stream_events`, so downstream consumers (`extract_conversation_text`, `_extract_tool_calls_from_events`) see the same flat schema regardless of source.
- Batch-mode fallback restored: `events.*` and `stdout.log` are looked up under `runs_dir/run_id/` when they don't exist per-case (Harbor writes them at the run root).
- All `read_text()` calls pass `encoding="utf-8", errors="replace"` — a single non-UTF-8 artifact no longer aborts a case.
- `input.yaml` is loaded into the `{{ inputs }}` Jinja variable; nested `dict`/`list` values render via `yaml.safe_dump`.
- Conversation fallback (stdout.log when events are empty) is now gated on `record["events"]` being empty — otherwise raw stream-json would be dumped into `{{ conversation }}`.
- `_extract_verifiable_evidence` (for `{{ evidence }}`) is derived entirely from `record["events"]` — no file discovery, no re-parse, no Claude-specific paths. Runner-agnostic: matches tool names (`Read`/`read_file`/`view`, `Write`/`write_file`/`edit`, `Bash`/`shell`/`execute`) and input keys (`file_path`/`path`/`file`, `command`/`cmd`) across Claude Code, opencode, codex, and responses-api.
- Shell-aware script extraction: `shlex.split` + filters for option flags (`--input=x.py`) and shell redirect targets (`> log.py`), so `./run.sh --input=x.py > log.py` records only `run.sh`.
- Evidence is extracted lazily (only when a template references `{{ evidence }}`) and cached back into `out["evidence"]`, avoiding repeat work across judges × samples.

### `agent_eval/harbor/run.py`, `agent_eval/evalhub/runner.py`

- Pass `reward_cfg` via the new `report.load_reward_cfg(config_path)` helper so containerized report rendering matches training too.

## Test plan

- [x] `python3 -m pytest tests/ -v` — unit tests pass locally
- [ ] `/eval-run` on an eval with a `reward:` section: verify the report's Reward column and Average match `build_reward` / Harbor's `reward.json`
- [ ] `/eval-run` in **batch** mode against a Harbor pod job: verify per-case `events` and `conversation` are populated (via the run-root fallback)
- [ ] Render report on a run mixing check gates, LLM judges, and both LLM and Python builtins: verify column grouping, headers, and color bands
- [ ] LLM judge template referencing `{{ evidence }}`: verify tool calls / files / scripts render correctly for a Claude Code run and remain accurate for a non-Claude runner
- [ ] Non-UTF-8 bytes in `stdout.log` / `events.jsonl` / `input.yaml`: verify the case still loads (replacement chars)

## Summary by CodeRabbit

* **New Features**
  * Added a “Per-Case Reward Overview” section to generated HTML reports, with a horizontally scrollable judge/reward table and an average row.
  * Expanded scoring context so judge prompts can use `inputs` and extracted `evidence` from agent run traces.

* **Bug Fixes**
  * Improved reward computation and rendering in reports, including applying reward configuration when available and clearly handling unscored cases.
  * Enhanced case record loading to better ingest event logs and populate conversation/evidence inputs reliably.

* **Documentation**
  * Updated Jinja2 template variable references for `inputs` and `evidence` in LLM judge prompts.


Sridhar1030:
@coderabbitai review

Sridhar1030:
@coderabbitai review

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,173 @@ def _colorise_hist(glyph):
...
+    for jn in gate_judges + llm_judges + other_judges:
+        html += f'<th class="rv-rotate">{_label(jn)}</th>'
+    html += '</tr>\n'
+
+    rewards = []
+    for case_id in sorted(per_case.keys()):
+        case_results = per_case[case_id]
+        if not isinstance(case_results, dict):
+            continue
+        grpo = case_results.get("grpo_reward", {})
```
astefanutti: **Reward column only works for `formula: weighted`, and it duplicates (and diverges from) the canonical reward engine.**

Two problems compound here:

1. `case_results.get("grpo_reward")` is the primary source, but **`grpo_reward` is never produced by the scoring pipeline** — grep shows the name exists only in this file. The real per-case reward is computed by `compose_reward()` / `compute_reward_from_config()` in `agent_eval/harbor/reward.py` and written to Harbor's `reward.json`, never into `summary["per_case"]`. So `reward_val` is `None` for every case and this falls through to the weighted branch.
2. The weighted branch only handles `formula: weighted`. For a single-`judge` reward, a Python-expression `formula`, the default (no `reward:` section), or any gated reward, **every Reward cell and the Average row render `—`** — i.e. the headline feature is blank for most configs. Even for `weighted` it ignores gate-zeroing and normalizes differently than the canonical function.

Rather than reimplementing the math here, import and call the canonical `compute_reward_from_config(per_case_judges, reward_cfg)` from `agent_eval.harbor.reward` (build the `RewardConfig` via the parser in `config.py`). That way the report can never disagree with the reward the harness actually trains on.

skills/eval-run/scripts/score.py
```diff
@@ -177,31 +269,78 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
...
+            if not events_path.exists():
+                events_path = runs_dir / run_id / "events.jsonl"
     if events_path.exists():
         try:
-            with open(events_path) as f:
-                record["events"] = json.load(f)
+            raw_text = events_path.read_text()
+            if events_path.suffix == ".jsonl":
+                record["events"] = [
+                    json.loads(line) for line in raw_text.splitlines() if line.strip()
```
astefanutti: **One malformed line discards *all* events.**

The `.jsonl` branch parses every line inside a single `try/except (json.JSONDecodeError, OSError)`, so a single non-JSON / truncated line (common at the tail of an interrupted Harbor pod transcript) raises `JSONDecodeError`, the outer handler sets `record["events"] = []`, and all valid events are thrown away — zeroing conversation extraction, tool-call extraction, and every event-based judge for that case.

Note `_extract_verifiable_evidence` (lines 94-97) already tolerates bad lines per-line; this path dropped that tolerance. Skip bad lines instead:

```suggestion
                record["events"] = []
                for line in raw_text.splitlines():
                    line = line.strip()
                    if not line:
                        continue
                    try:
                        record["events"].append(json.loads(line))
                    except json.JSONDecodeError:
                        continue
```

skills/eval-run/scripts/score.py
```diff
@@ -53,6 +53,98 @@ def _resolve_under(root: Path, candidate: Path) -> Path:
...
+    tools = collections.Counter()
+    skills_invoked = []
+    scripts_run = set()
+    arch_files_read = set()
+    input_read = False
+    artifacts_written = set()
+    total_turns = 0
+    cost_usd = 0.0
+
+    key_scripts = [
```
astefanutti: **Project-specific hardcoding inside the generic framework.**

`key_scripts` (`fetch-architecture-context.sh`, `bootstrap-assess-rfe.sh`, `prep_assess.py`, `next_rfe_id.py`, …), the `.context/architecture` path, and the `artifacts/` prefix are all **rfe-creator** internals — they exist nowhere else in `agent_eval/`. CLAUDE.md's first design principle is *"generic evaluation framework … schema-driven, no hardcoded field names."*

For any other skill project the `{{ evidence }}` block becomes misleading noise (`fetch-architecture-context.sh: NO`, `Architecture context files read: NONE`, `input.yaml read: NO`) injected into every LLM judge prompt, biasing judges against correct runs. This belongs in the project's `eval.yaml` (e.g. judge `arguments` listing the scripts/paths to watch), not baked into the shared engine. Keep the generic parts (turn count, cost, tool counter, skills invoked) and drive the rest from config.

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,173 @@ def _colorise_hist(glyph):
...
+                sr = reward_cfg.get("score_range", [1, 5])
+                if not isinstance(sr, list) or len(sr) < 2:
+                    sr = [1, 5]
+                raw_list = reward_cfg.get("raw", [])
+                total = 0.0
+                wsum = 0.0
+                for jn, w in weights.items():
+                    jentry = case_results.get(jn, {})
+                    v = jentry.get("value") if isinstance(jentry, dict) else None
+                    if v is not None and isinstance(v, (int, float)):
```
astefanutti: **Booleans are treated as numeric scores** (`isinstance(True, int)` is `True` in Python).

If a boolean gate judge is listed in `reward.weights`, `passes=True` normalizes to `(1 - 1)/(5 - 1) = 0.0` and `passes=False` to a clamped `0.0` — so a *passing* gate contributes `0.0` to the reward instead of `1.0`, silently corrupting the displayed value. Every other classifier in this PR checks `isinstance(v, bool)` before the numeric branch (see `_cell` at 2007-2010, and 1985-1988); this branch should too:

```suggestion
                    if isinstance(v, bool):
                        norm = 1.0 if v else 0.0
                        total += float(w) * norm
                        wsum += float(w)
                    elif isinstance(v, (int, float)):
```
(This is another reason to just call `compute_reward_from_config`, which already handles this via `_judge_value`.)

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,173 @@ def _colorise_hist(glyph):
...
+            continue
+        jtype = j.get("type", "")
+        if jtype == "check" and j.get("check", ""):
+            gate_judges.append(name)
+        elif jtype == "llm" or j.get("prompt") or j.get("prompt_file"):
+            llm_judges.append(name)
+        else:
+            other_judges.append(name)
+
+    if not gate_judges and not llm_judges:
```
astefanutti: **Duplicate judge columns when no configured judge is a gate or LLM judge.**

The config loop above classifies each judge; `builtin`/`module` judges (no `check`, no `prompt`/`prompt_file`) land in `other_judges`. If *all* configured judges are that kind (e.g. `efficiency` + a builtin `quality`), then `gate_judges` and `llm_judges` are both empty, so this `if not gate_judges and not llm_judges:` fallback runs and **re-adds the same judge names** from `per_case` into `gate/llm/other`. The judge then appears twice in the header (`gate_judges + llm_judges + other_judges`) and twice in every row.

Guard the fallback against names already classified, or make it an `elif`/`else` that only runs when the config loop produced nothing at all.

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,173 @@ def _colorise_hist(glyph):
...
+
+    def _cell(val_entry):
+        if not isinstance(val_entry, dict):
+            return '<span class="skip">\u2014</span>'
+        v = val_entry.get("value")
+        if v is True:
+            return '<span class="pass">PASS</span>'
+        if v is False:
+            return '<span class="fail">FAIL</span>'
+        if isinstance(v, float):
```
astefanutti: **Hardcoded 1–5 thresholds mis-color judges that aren't on a 1–5 scale.**

A judge in the *Other* group that emits a 0–1 ratio — e.g. `efficiency` at `0.9` — hits this float branch and, since `0.9 < 2.5`, renders red **FAIL** despite being a near-perfect score. The framework supports a configurable `score_range`, so these fixed cutoffs are wrong for anything but a 1–5 LLM judge.

Also, the int branch (2018-2023) uses different boundaries than the float branch (`>=3` vs `>=2.5` for warn), so the same semantic score colors differently depending on whether the judge returned `3` or `3.0`. Consider deriving bands from the judge's `score_range`, and only applying them to the LLM group.

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,173 @@ def _colorise_hist(glyph):
...
+            other_judges.append(name)
+
+    if not gate_judges and not llm_judges:
+        all_judge_names = set()
+        for case_results in per_case.values():
+            if isinstance(case_results, dict):
+                for jn, jv in case_results.items():
+                    if jn != "grpo_reward" and isinstance(jv, dict):
+                        all_judge_names.add(jn)
+        for jn in sorted(all_judge_names):
```
astefanutti: **Auto-detection samples the *first* case a judge appears in — a skipped first case misclassifies it.**

If a judge has an `if:` condition that skips it for the alphabetically-first case, `per_case[first][judge]["value"]` is `None`, so neither the `bool` nor `(int, float)` check matches and the judge is filed under *Other* — even though it produces a 1–5 score in the cases where it *did* run. It then renders under the wrong header with the wrong coloring. Scan cases until you find a non-`None` value before classifying.

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,173 @@ def _colorise_hist(glyph):
...
+                            norm = max(0.0, min(1.0, (float(v) - sr[0]) / span)) if span else 0.0
+                        total += float(w) * norm
+                        wsum += float(w)
+                reward_val = total / wsum if wsum > 0 else None
+        html += f'<tr><td>{_esc(case_id)}</td>'
+        html += f'<td>{_reward_cell(reward_val)}</td>'
+        for jn in gate_judges + llm_judges + other_judges:
+            html += f'<td>{_cell(case_results.get(jn, {}))}</td>'
+        html += '</tr>\n'
+        try:
```
astefanutti: **Average silently computed over a subset of cases.**

`float(reward_val)` throws `TypeError` for `None` and is swallowed, so cases with no reward are excluded from `rewards`. The row then reads `Average (N cases)` where `N` is smaller than the number of rows rendered above — a partial-failure run (several `—` rewards) shows an inflated mean without making it obvious those cases were dropped. Consider counting missing rewards as `0.0`, or label it clearly (e.g. `Average (6 of 10 scored)`).

skills/eval-run/scripts/score.py
```diff
@@ -177,31 +269,78 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
...
         from agent_eval.events import extract_conversation_text
         record["conversation"] = extract_conversation_text(record["events"])
     else:
         record["conversation"] = ""
 
+    # Fallback: build conversation from stdout.log if events are empty
+    # (Harbor pods write agent output to stdout.log, not events.json)
+    if not record["conversation"]:
+        stdout_path = case_dir / "stdout.log"
+        if stdout_path.exists():
```
astefanutti, line 383: **`stdout.log` is interpreted two contradictory ways for the same case.**

This conversation fallback reads `stdout.log` as **plain text**, while `_extract_verifiable_evidence` (line 64) reads the *same* file as **line-delimited stream-json**. For the local `claude-code` runner that's fine (it always writes stream-json), but the `cli_runner` and other agents write arbitrary plain-text stdout — there, `{{ conversation }}` gets the real transcript while `{{ evidence }}` reports `Tool calls: none / Skills invoked: none`, giving a judge self-contradictory context. Gate the evidence parse on the runner/format, or detect stream-json before assuming it.

skills/eval-run/scripts/score.py
```diff
@@ -177,31 +269,78 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
...
+            try:
+                record["conversation"] = stdout_path.read_text()
+            except OSError:
+                pass
+
+    # --- Verifiable evidence from agent trace ---
+    record["evidence"] = _extract_verifiable_evidence(case_dir, record)
```
astefanutti: **Evidence is extracted for every case unconditionally, even when nothing consumes it.**

`_extract_verifiable_evidence` opens and JSON-parses the entire trace on every `load_case_record` call — including the two calls per case inside `compare_runs` (pairwise) — regardless of whether any judge template references `{{ evidence }}`. On large stream-json traces across many cases this is pure wasted I/O/CPU. Compute it lazily (only when a judge/template uses `evidence`), and reuse the already-parsed `record["events"]` instead of re-reading the file from disk.

skills/eval-run/scripts/score.py
```diff
@@ -767,6 +913,10 @@ def _score_case(case_dir):
         record = load_case_record(case_dir, config, run_id=run_id)
         case_results = {}
         for name, scorer, condition, judge_type, judge_samples in judges:
+            import copy
+            record["prior_judges"] = {
```
astefanutti: **`prior_judges`: quadratic deepcopy, in-loop import, no consumer, and a description mismatch.**

- `import copy` runs on every judge iteration — hoist to module top.
- `record["prior_judges"]` is rebuilt as a full `deepcopy` of all accumulated results before *each* judge → O(J²) deep copies per case, for no benefit (judges receive `outputs` read-only). A shallow snapshot, or copying only the newly-added entry, suffices.
- Nothing in the repo reads `prior_judges` (grep-confirmed), so this is currently dead infrastructure.
- The PR description says downstream judges get `outputs["judge_<name>"]`, but the code writes `outputs["prior_judges"][<name>]`. Please reconcile the two (and ideally land the one consumer + a test, otherwise drop it).

skills/eval-run/scripts/score.py
```diff
@@ -177,31 +269,78 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
...
     # --- Events (structured event stream) ---
+    # Support both events.json (JSON array) and events.jsonl (one JSON per line,
+    # as produced by Claude Code session transcripts in Harbor pods).
     events_path = case_dir / "events.json"
+    if not events_path.exists():
+        events_path = case_dir / "events.jsonl"
     if not events_path.exists() and run_id and runs_dir:
-        events_path = runs_dir / run_id / "events.json"
+        if ".." in str(run_id) or Path(run_id).is_absolute():
+            print(f"  Warning: rejecting unsafe run_id: {run_id}",
```
astefanutti: **This CWE-22 guard protects a code path that never executes, and duplicates an existing helper.**

`load_case_record` is only ever called with `run_id` and **without** `runs_dir` (see `_score_case` at line 913 and the two `compare_runs` calls), so `runs_dir` is always `None` and this `if not events_path.exists() and run_id and runs_dir:` block — plus the `.jsonl` fallback and the matching stdout/stderr fallbacks below — is dead. If you intend these run-dir fallbacks to work, pass `runs_dir` through; if not, they can go. Separately, the hand-rolled `".." in str(run_id) or is_absolute()` check duplicates `_resolve_under()` (line 44), which also catches symlink / `a/../../b` escapes the string check misses.

skills/eval-run/scripts/score.py
```diff
@@ -53,6 +53,98 @@ def _resolve_under(root: Path, candidate: Path) -> Path:
...
+    """
+    import collections
+
+    trace_path = Path(case_dir) / "stdout.log"
+    if not trace_path.exists():
+        agent_trace = Path(case_dir) / "agent" / "claude-code.txt"
+        if agent_trace.exists():
+            trace_path = agent_trace
+        else:
+            for p in Path(case_dir).glob("**/*claude*.txt"):
```
astefanutti: **Non-deterministic trace selection.**

`for p in Path(case_dir).glob("**/*claude*.txt"): trace_path = p; break` takes the first match from an *unordered* generator. A case dir with more than one match (e.g. `agent/claude-code.txt` plus a subagent transcript) picks a filesystem-dependent file, so evidence differs between a dev laptop and CI, and may parse a subagent transcript instead of the main one. Use `sorted(...)` and/or a specific expected filename.

skills/eval-run/scripts/score.py
```diff
@@ -177,31 +269,78 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
...
+    # --- Input data (from input.yaml in case directory or dataset) ---
+    record["input"] = ""
+    input_yaml = case_dir / "input.yaml"
+    if not input_yaml.exists() and config.dataset.path:
+        dataset_root = config.resolve_path(config.dataset.path).resolve()
+        input_yaml = dataset_root / case_id / "input.yaml"
+    if input_yaml.exists():
+        try:
+            raw = yaml.safe_load(input_yaml.read_text()) or {}
+            if isinstance(raw, dict):
```
astefanutti, line 353: **Nested/list `input.yaml` values render as Python `repr` into judge prompts.**

`f"**{key}**: {val}"` stringifies structured values with Python syntax — `tags: [a, b]` becomes `**tags**: ['a', 'b']` and a nested mapping becomes `{'k': 'v'}`. That Python-flavored formatting (quotes, brackets) leaks into `{{ input }}` LLM judge prompts and reads worse than the source YAML. Consider `yaml.safe_dump` for non-scalar values, or rendering nested structures recursively.

astefanutti (commented):
@Sridhar1030 thanks, that's great! I've left some comments.

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,180 @@ def _colorise_hist(glyph):
...
+        pass
+
+    judges_cfg = config.get("judges", [])
+    classified = set()
+    gate_judges = []
+    llm_judges = []
+    other_judges = []
+    for j in judges_cfg:
+        name = j.get("name", "")
+        if name in ("grpo_reward",):
```
astefanutti: **Dead `grpo_reward` hardcode.** `grpo_reward` is produced nowhere in the codebase (grep-confirmed: only these two guards + a historical CHANGELOG line; the `reward:` section is not a judge and never appears in `case_results`). These skip-guards (here and line 1986) filter a key that can never exist — inert, and misleading since they imply `grpo_reward` is a live convention. It was removed when the `reward:` config section replaced it. Please delete both.

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,180 @@ def _colorise_hist(glyph):
...
+            continue
+        classified.add(name)
+        jtype = j.get("type", "")
```
astefanutti: **Judge classification is broken and duplicates an existing source of truth.**

- Line 1974 keys on `j.get("type")`, but eval.yaml judges have **no `type` field** (`JudgeConfig` in `config.py` has none; the discriminator is which of `check:`/`prompt:`/`builtin:` is set). So `jtype` is always `""`, `jtype == "check"` never matches, and an inline **`check` gate judge renders under "Other" instead of the Gate group**.
- A **`builtin` LLM judge** (1–5, no `prompt`/`prompt_file`) also falls to `other_judges`, so `_cell` colors it on the 0–1 scale (`>=0.7` pass) — every 1–5 score, even a failing **1**, renders green PASS.
- Both are avoidable: the authoritative `judge_type` (check/llm/builtin/code) is already stamped into each per-case result by `score.py`, and **you already read it in this file at line 1328** (`(first_case.get(name) or {}).get("judge_type")`). Reuse it here instead of re-deriving — one change fixes both misclassifications.

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,180 @@ def _colorise_hist(glyph):
...
+            if sample_val is None:
+                other_judges.append(jn)
+            elif isinstance(sample_val, bool):
+                gate_judges.append(jn)
+            elif isinstance(sample_val, (int, float)):
+                llm_judges.append(jn)
+            else:
+                other_judges.append(jn)
+
+    has_efficiency = any(
```
astefanutti: **Hardcoded judge name in generic code.** The `efficiency` judge is special-cased by literal name (detected here, force-appended, then reordered to the end of `other_judges`). Any project that renames it loses the handling, and an unrelated judge named `efficiency` gets forced into "Other". Grouping should follow judge *type* (see the classification comment above), not a magic string — contradicts the CLAUDE.md "no hardcoded field names" principle.

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,180 @@ def _colorise_hist(glyph):
...
+        other_judges.append("efficiency")
+    other_judges = [j for j in other_judges if j != "efficiency"] + (
+        ["efficiency"] if "efficiency" in other_judges else [])
+
+    llm_judge_set = set(llm_judges)
+
+    def _label(name):
+        return _esc(name.replace("_", " ").title())
+
+    def _cell(val_entry, judge_name=""):
```
astefanutti, line 2048: **Color bands ignore the configurable `score_range`.** `_cell` assumes LLM judges are 1–5 and other numerics are 0–1; `_reward_cell` (line 2047) hardcodes 0.40/0.30 cutoffs. The framework supports a configurable `score_range` (and `reward.raw` for [0,1] judges), so these fixed cutoffs mis-color any judge on a different scale — e.g. a 0–10 LLM judge shows every score ≥4 as green. Derive the bands from each judge's `score_range`.

skills/eval-run/scripts/report.py
```diff
@@ -1926,6 +1945,180 @@ def _colorise_hist(glyph):
...
+    for case_id in sorted(per_case.keys()):
+        case_results = per_case[case_id]
+        if not isinstance(case_results, dict):
+            continue
+        total_cases += 1
+        # Use canonical reward engine when available
+        reward_val = None
+        if _compose_reward is not None:
+            try:
+                reward_val, _ = _compose_reward(case_results)
```
astefanutti: **Reward is computed without the config, so it diverges from the trained reward.** `compose_reward(case_results)` is called with no `reward_cfg`, so it always uses the *default* resolution (bool gates + numerics averaged on a fixed [1,5] range). For any eval with a `reward:` section (single-`judge`/`weighted`/`formula`/gate) this column and the Average differ from what `build_reward` trains on. `report.py` currently only has the raw YAML dict (`_load_yaml`, line 2654), so it can't pass a `RewardConfig` today — thread `EvalConfig.from_yaml(args.config).reward` through `generate_report` to here.

Two concrete artifacts of the default path (both fixed by passing `reward_cfg`):
- A case where **all judges are skipped/None** yields `reward = 1.0` (rendered green 1.0000) and is counted in the Average — inflating the mean for unscored cases.
- The default path gates to **0.0 on *any* False boolean judge** (not just configured gates), so a strong case with one incidental False bool shows 0.0000, contradicting its own green per-judge cells.

skills/eval-run/scripts/score.py
```diff
@@ -53,6 +53,98 @@ def _resolve_under(root: Path, candidate: Path) -> Path:
 # Case record loading — reads all files, no schema interpretation
 # ---------------------------------------------------------------------------
 
+def _extract_verifiable_evidence(case_dir, record):
```
astefanutti: **Evidence re-reads and re-parses the trace from disk instead of reusing what's already parsed.** `load_case_record` already parsed the transcript into `record["events"]`, and it is passed in here as `record` — but the argument is never used; this function re-opens `stdout.log`/`*claude*.txt` and parses it again. Derive the evidence from `record["events"]` instead (which is already normalized across runners — this also removes the claude-specific schema assumptions and the format-sniffing below).

skills/eval-run/scripts/score.py
```diff
@@ -53,6 +53,98 @@ def _resolve_under(root: Path, candidate: Path) -> Path:
...
+    is_stream_json = False
+    if trace_path.exists():
+        try:
+            with trace_path.open() as fh:
+                for probe in fh:
+                    probe = probe.strip()
+                    if not probe:
+                        continue
+                    try:
+                        json.loads(probe)
```
astefanutti: **Format probe misfires on a scalar/banner first line.** `json.loads(probe)` succeeds on bare scalars (`json.loads("42"/"true"/"null")`), so a plain-text trace starting with a lone token is misdetected as stream-json → all-`none` evidence. Conversely, a real transcript whose first line is a non-JSON banner is judged plain-text (the `break` probes only the first non-empty line) → empty evidence. Deriving from `record["events"]` (see the function-level comment) avoids this entirely.

skills/eval-run/scripts/score.py
```diff
@@ -53,6 +53,98 @@ def _resolve_under(root: Path, candidate: Path) -> Path:
...
+                                fp = inp.get("file_path", "")
+                                if fp:
+                                    files_read.add(fp.split("/")[-1])
+                            elif name == "Write":
+                                fp = inp.get("file_path", "")
+                                if fp:
+                                    files_written.add(fp.split("/")[-1])
+                    elif d.get("type") == "result":
+                        total_turns = d.get("num_turns", 0)
+                        cost_usd = d.get("total_cost_usd", 0.0)
```
astefanutti: **`total_cost_usd: null` crashes evidence extraction.** `d.get("total_cost_usd", 0.0)` returns `None` when the key is *present but null* (e.g. an aborted result event) — `.get` only substitutes the default when the key is absent. The later `f"Cost: ${cost_usd:.2f}"` (line 138) then raises `TypeError`, and it sits outside the `except OSError`, so the judge for that case errors instead of scoring. Use `cost_usd = d.get("total_cost_usd") or 0.0`.

skills/eval-run/scripts/score.py
```diff
@@ -177,31 +269,79 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
                 pass
 
     # --- Events (structured event stream) ---
+    # Support both events.json (JSON array) and events.jsonl (one JSON per line,
+    # as produced by Claude Code session transcripts in Harbor pods).
     events_path = case_dir / "events.json"
-    if not events_path.exists() and run_id and runs_dir:
-        events_path = runs_dir / run_id / "events.json"
+    if not events_path.exists():
+        events_path = case_dir / "events.jsonl"
```
astefanutti, line 309: **Batch mode now gets empty `events` and `conversation`.** The removed fallback (`runs_dir/run_id/events.json`) was actually reachable — `runs_dir` is defaulted at line 157 and the scorer passes `run_id` (line 920). In **batch** mode `collect.py`/`execute.py` write `events.json` and `stdout.log` at the **run root**, never per-case, so `case_dir/events.{json,jsonl}` (here) and `case_dir/stdout.log` (line 334) don't exist → every batch case gets `record["events"]=[]` and `record["conversation"]=""`, silently zeroing all event-based judges and `{{ conversation }}`. Restore a run-root fallback for the batch layout (case mode is unaffected).

skills/eval-run/scripts/score.py
```diff
@@ -177,31 +269,79 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
...
     if events_path.exists():
         try:
-            with open(events_path) as f:
-                record["events"] = json.load(f)
+            raw_text = events_path.read_text()
```
astefanutti: **`UnicodeDecodeError` is uncaught in three `read_text()` sites.** Here (events), line 311 (`input.yaml`), and line 337 (conversation fallback) call `read_text()` without `encoding=`. `UnicodeDecodeError` subclasses `ValueError`, not `OSError`/`JSONDecodeError`, so none of the surrounding handlers catch it — a non-UTF-8 Harbor artifact crashes `load_case_record` for the whole case. (The file-artifact reads elsewhere in this function correctly catch `UnicodeDecodeError`.) Pass `encoding="utf-8", errors="replace"` or catch it.

skills/eval-run/scripts/score.py
```diff
@@ -177,31 +269,79 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
...
+            if events_path.suffix == ".jsonl":
+                record["events"] = []
+                for line in raw_text.splitlines():
+                    line = line.strip()
+                    if not line:
+                        continue
+                    try:
+                        record["events"].append(json.loads(line))
```
astefanutti: **Two problems on the new `.jsonl` path:**

1. **Schema mismatch (silent empty conversation + zero tool_calls).** These lines are appended as **raw** Claude transcript objects (per the comment, "Claude Code session transcripts in Harbor pods" — text/tools nested under `message.content[]`). But the consumers of `record["events"]` read the **flat** schema: `extract_conversation_text` reads `event.get("text")` (`events.py:346`) and `_extract_tool_calls_from_events` reads `event.get("tools")` (line 409). So for exactly the Harbor case this branch targets, `conversation` is empty and tool-output judges see zero calls. Normalize raw lines to the flat schema on load, or route them through the raw parser.
2. **Crash on a scalar line.** Only `JSONDecodeError` is caught, so a valid-but-scalar line (`null`→`None`, `42`, `"x"`) is appended as a non-dict; the unguarded `event.get(...)` in the consumers above then raises `AttributeError`, failing the case. Guard with `isinstance(obj, dict)`.

skills/eval-run/scripts/score.py
```diff
@@ -177,31 +269,79 @@ def load_case_record(case_dir, config, run_id=None, runs_dir=None):
...
     else:
         record["conversation"] = ""
 
+    # Fallback: build conversation from stdout.log if events are empty
+    # (Harbor pods write agent output to stdout.log, not events.json)
+    if not record["conversation"]:
+        stdout_path = case_dir / "stdout.log"
+        if stdout_path.exists():
+            try:
+                record["conversation"] = stdout_path.read_text()
```
astefanutti: **Conversation fallback can inject raw stream-json into `{{ conversation }}`.** The guard is `if not record["conversation"]` (empty *string*), not `if not record["events"]`. So when events parsed but `extract_conversation_text` returned `""` (e.g. all content nested under subagents), the raw `stdout.log` — a wall of stream-json for CC runs — is dumped into the judge prompt. Gate the fallback on events being empty, or extract readable text rather than dumping raw stdout. (This line is also one of the uncaught-`UnicodeDecodeError` sites noted above.)

skills/eval-run/scripts/score.py
```diff
@@ -364,12 +504,25 @@ def _render_jinja2_template(template_text, arguments, outputs):
...
         conversation = extract_conversation_text(out["events"])
 
+    # Pre-render input data for {{ input }}
+    input_text = out.get("input", "")
+
     template = env.from_string(template_text)
+
+    # Lazy evidence: only extract if template uses {{ evidence }}
+    evidence_text = out.get("evidence", "")
+    if not evidence_text and "evidence" in template_text:
```
astefanutti: **Evidence is re-extracted once per judge × per sample.** The result is never written back to `out["evidence"]`, and `score_cases` calls the scorer (→ this render) once per judge and once per sample. With J evidence-using judges and S samples the trace is re-opened + re-parsed **J×S times per case**. Cache it (`out["evidence"] = evidence_text`) after first extraction — combined with reusing `record["events"]` (see the `_extract_verifiable_evidence` comment) this becomes ~free. Note the trigger is also a loose substring (`"evidence" in template_text`), so prose merely mentioning the word re-triggers extraction.

astefanutti (commented):
## Follow-up deep review of the updates

9-angle finder pass + per-finding verification + a gap sweep over the two changed files. **15 findings, posted inline below** (grouped by root cause into 13 threads), ranked most-severe first within each category:

- **Silent wrong output:** reward divergence from the trained reward; batch-mode empty `events`/`conversation`; `.jsonl` raw-vs-flat schema mismatch; default-reward `1.0` for skipped cases / `0.0` on any false bool; check-gate + builtin-LLM misclassification.
- **Crashes:** `total_cost_usd: null` → `TypeError`; non-UTF-8 `read_text()`; scalar `.jsonl` line → `AttributeError`.
- **Cleanup/altitude:** dead `grpo_reward` hardcode; hardcoded `efficiency`; color bands ignore `score_range`; re-parsing traces (reuse `record["events"]`); re-derived classification (reuse `judge_type`).

Several share a root cause and collapse into single fixes: **#classification + #reuse** (use the per-case `judge_type` already read at report.py:1328), and **#reward** (thread `reward_cfg` from `EvalConfig`).

**Also verified but cut for brevity (lower severity):** `{{ input }}` undocumented + no `StrictUndefined`; scalar `input.yaml` values leak Python reprs (`no`→`False`, `null`→`None`); `<td colspan="0">` when there are no judge columns; NaN/inf reward rendered as colored text; `files_read/written` keyed on basename only; `Skill` calls missing a `skill` input leak a literal `?`; `.sh`/`.py` token heuristic misreads flags as scripts; `except Exception: pass` / silent `ImportError` around the reward hide real failures.

**Refuted (not a bug):** batch case-dir names *do* equal dataset `case_id`s, so `{{ input }}` loads the correct case.

astefanutti (commented):
@Sridhar1030 I've run a second review cycle and left some comments.

Sridhar1030:
@astefanutti Addressed all 15 findings from the second review cycle

astefanutti (approved):
Thanks @Sridhar1030!

PR #147, opened 2026-07-03
feat(eval-analyze): add --assess flag to score skills for eval-worthiness

### LINKED ISSUE #144: /eval-analyze: Add batch assessment mode to identify which skills need evals
## Context

  When a project has many skills (e.g., 54 across 17 plugins), deciding which ones
  warrant eval investment is the first question teams face. Currently `/eval-analyze`
  asks the user to pick a single skill when multiple are found (Step 1). There's no
  way to assess all skills at once and get a recommendation.

  ## Proposal

  When `/eval-analyze` finds multiple skills and no `--skill` flag is provided, offer
  an additional strategy: **"Assess all skills and recommend which ones need evals."**

  This mode would:
  1. Run the skill analysis (Step 3) on each skill — read SKILL.md, check complexity,
     output type, tools used
  2. For each skill, determine how many meaningful judges it would generate
     (deterministic checks vs LLM judges)
  3. Produce a summary report:

  Skill Assessment Report
  ━━━━━━━━━━━━━━━━━━━━━━
  cluster-diagnostic      8 judges (5 check, 3 LLM)  → RECOMMENDED
  threat-model:tnf        6 judges (4 check, 2 LLM)  → RECOMMENDED
  hello-world             1 judge  (1 check, 0 LLM)  → SKIP
  apply-tide-squash-label 0 judges                    → SKIP

  The signal: skills that produce many meaningful judges (especially deterministic
  checks) are eval-worthy. Skills that produce only trivial checks are not worth
  the investment.

  ## Why

  - Teams with many skills need a data-driven way to prioritize eval coverage
  - Running `/eval-analyze` one skill at a time and manually comparing is tedious
  - The skill analysis logic already exists in Step 3 — this extends it to a
    batch summary
  - Discussed in the working group: https://github.com/opendatahub-io/agent-eval-harness — first use case at this scale (54 skills)

  ## Alternatives considered

  - Writing a wrapper script that calls `/eval-analyze --skill <name>` in a loop —
    works but duplicates logic and doesn't benefit the broader community
  - Manual categorization based on reading SKILL.md files — subjective and
    not repeatable

kasturinarra:
## Summary

- Adds `assess_skills.py` which reads all SKILL.md files, extracts eval-relevant metadata (tools, line count, complexity signals), and scores each skill to recommend whether evals would add value
- Updates `SKILL.md` to route `--assess` to the new script and offer batch assessment as an interactive option when multiple skills are found
- Categories: RECOMMENDED (score>=4), OPTIONAL (score>=2), SKIP, EXISTS (already has eval)

Closes #144

## Test plan

- [x] `python3 assess_skills.py` produces correct human-readable report
- [x] `python3 assess_skills.py --json` produces valid JSON output
- [x] Malformed SKILL.md (no closing `---`) handled gracefully (no crash)
- [x] Complex skills with "List" in description not falsely flagged as thin wrappers
- [ ] Run `/eval-analyze --assess` end-to-end on a project with multiple skills

## Summary by CodeRabbit

* **New Features**
  * Added an `--assess` option to run a batch assessment of available skills and display grouped recommendations (human-readable or JSON).
* **Behavior Changes**
  * Assessment mode now takes precedence over `--skill`, skipping the normal flow and stopping after recommendations.
  * Without `--assess`, skill selection now supports `--skill`, otherwise lists skills and prompts when multiple matches exist.
* **Bug Fixes**
  * Improved skill discovery by only processing valid skill files within the project directory (invalid/out-of-scope paths are skipped).
* **Documentation**
  * Added an assessment prompt specification describing recommendation logic and report format.

### DIFF
--- a/skills/eval-analyze/SKILL.md
+++ b/skills/eval-analyze/SKILL.md
@@ -42,6 +42,12 @@ If `--config` provided, use that path. Otherwise run `python3 ${CLAUDE_SKILL_DIR
 
 Set the resolved config path as `<config>` for all subsequent steps. Set `<eval_md_path>` to the same directory as `<config>`, with filename `eval.md`.
 
+**Modes**: 
+- `--skill my-skill` → skill-based eval (`execution.skill`, case/batch mode)
+- `--prompt examples/openshift-agentic-docs.md` → prompt-based eval (`execution.prompt`, case mode, tests agent capabilities)
+- `--assess` → batch assessment of all skills (no eval.yaml generated)
+- See `examples/` for domain-specific analysis prompt recipes
+
 ## Batch Assessment
 
 If `--assess` was provided (even with `--skill`, which is ignored in this mode):
@@ -51,11 +57,6 @@ python3 ${CLAUDE_SKILL_DIR}/scripts/assess_skills.py --json
 ```
 
 If empty, report no skills found and stop. Otherwise, read `${CLAUDE_SKILL_DIR}/prompts/assess-skills.md` and follow its instructions to classify each skill and display the grouped report. Offer to run `/eval-analyze --skill <name>` on RECOMMENDED skills. **Stop here — do not proceed to Step 1.**
-
-**Modes**: 
-- `--skill my-skill` → skill-based eval (`execution.skill`, case/batch mode)
-- `--prompt examples/openshift-agentic-docs.md` → prompt-based eval (`execution.prompt`, case mode, tests agent capabilities)
-- See `examples/` for domain-specific analysis prompt recipes
 
 ## Step 1: Determine Analysis Mode
 
@@ -79,7 +80,7 @@ If empty, report no skills found and stop. Otherwise, read `${CLAUDE_SKILL_DIR}/
 python3 ${CLAUDE_SKILL_DIR}/scripts/find_skills.py --name <skill>
 ```
 
-**If neither `--skill` nor `--assess` was provided**, list all project skills:
+**If `--skill` was not provided**, list all project skills:
 
 ```bash
 python3 ${CLAUDE_SKILL_DIR}/scripts/find_skills.py
--- a/skills/eval-analyze/SKILL.md
+++ b/skills/eval-analyze/SKILL.md
@@ -20,6 +20,7 @@ The core principle: **observe, don't assume**. Every field name, file pattern, a
 | `--config <path>` | no | auto-discover | Output path for the config |
 | `--prompt <path>` | no | none | Custom analysis prompt (for non-skill evals) |
 | `--update` | no | false | Fill in missing sections only, preserve user edits |
+| `--assess` | no | false | Assess all skills and recommend which ones need evals |
 
 ```bash
 mkdir -p tmp
@@ -58,19 +59,38 @@ Set the resolved config path as `<config>` for all subsequent steps. Set `<eval_
 
 ### Step 2: Find the Target Skill
 
-If `--skill` was provided, locate its SKILL.md:
+**If `--assess` was provided**, skip single-skill selection and run the batch assessment instead:
+
+```bash
+python3 ${CLAUDE_SKILL_DIR}/scripts/assess_skills.py
+```
+
+This reads all SKILL.md files in the project, extracts eval-relevant metadata (allowed tools, line count, complexity signals), and scores each skill to recommend whether evals would add value. Display the report to the user and offer to run full `/eval-analyze` on any of the recommended skills. Stop here — do not proceed to Step 2.
+
+**If `--skill` was provided**, locate its SKILL.md:
 
 ```bash
 python3 ${CLAUDE_SKILL_DIR}/scripts/find_skills.py --name <skill>
 ```
 
-If not provided, list all project skills:
+**If neither `--skill` nor `--assess` was provided**, list all project skills:
 
 ```bash
 python3 ${CLAUDE_SKILL_DIR}/scripts/find_skills.py
 ```
 
-This reads `.claude-plugin/plugin.json` for custom skill paths, falls back to `.claude/skills/` and `skills/`, and excludes eval harness skills. If only one skill is found, use it automatically. If multiple, ask the user which to analyze. If none are found, tell the user — they may need to check their skill directory paths or create a skill first.
+This reads `.claude-plugin/plugin.json` for custom skill paths, falls back to `.claude/skills/` and `skills/`, and excludes eval harness skills. If only one skill is found, use it automatically. If none are found, tell the user — they may need to check their skill directory paths or create a skill first.
+
+If multiple skills are found, ask the user which strategy to use via AskUserQuestion:
+
+- **"Analyze a specific skill"** — list the skills and let the user pick one (current behavior)
+- **"Assess all skills — recommend which ones need evals"** — run the batch assessment:
+
+```bash
+python3 ${CLAUDE_SKILL_DIR}/scripts/assess_skills.py
+```
+
+Display the assessment report and offer to run full `/eval-analyze` on any of the recommended skills. If the user picks a skill to analyze, continue to Step 2. Otherwise stop here.
 
 **If `--update` and eval.yaml already has a `skill` field**: use that skill. If `--skill` is also provided and differs, ask the user which they mean — don't silently overwrite.
 
--- a/skills/eval-analyze/SKILL.md
+++ b/skills/eval-analyze/SKILL.md
@@ -22,6 +22,10 @@ The core principle: **observe, don't assume**. Every field name, file pattern, a
 | `--update` | no | false | Fill in missing sections only, preserve user edits |
 | `--assess` | no | false | Assess all skills and recommend which ones need evals |
 
+**If `$ARGUMENTS` contains `--assess`**: skip the rest of Step 0 and Step 1. Go directly to **Batch Assessment** (between Step 0 and Step 1). Do not run `state.py init`, do not discover configs, do not look for a single skill.
+
+Otherwise, proceed with the normal flow:
+
 ```bash
 mkdir -p tmp
 python3 ${CLAUDE_SKILL_DIR}/scripts/agent_eval/state.py init tmp/analyze-config.yaml \
@@ -37,6 +41,16 @@ If `--config` provided, use that path. Otherwise run `python3 ${CLAUDE_SKILL_DIR
 - **`--config` provided**: use the explicit path, bypass layout logic
 
 Set the resolved config path as `<config>` for all subsequent steps. Set `<eval_md_path>` to the same directory as `<config>`, with filename `eval.md`.
+
+## Batch Assessment
+
+If `--assess` was provided (even with `--skill`, which is ignored in this mode):
+
+```bash
+python3 ${CLAUDE_SKILL_DIR}/scripts/assess_skills.py --json
+```
+
+If empty, report no skills found and stop. Otherwise, read `${CLAUDE_SKILL_DIR}/prompts/assess-skills.md` and follow its instructions to classify each skill and display the grouped report. Offer to run `/eval-analyze --skill <name>` on RECOMMENDED skills. **Stop here — do not proceed to Step 1.**
 
 **Modes**: 
 - `--skill my-skill` → skill-based eval (`execution.skill`, case/batch mode)
@@ -58,28 +72,6 @@ Set the resolved config path as `<config>` for all subsequent steps. Set `<eval_
 ## SKILL ANALYSIS (Default)
 
 ### Step 2: Find the Target Skill
-
-**If both `--assess` and `--skill` are provided**, warn the user that `--assess` performs batch assessment of all skills and `--skill` is ignored for this mode. Proceed with `--assess`.
-
-**If `--assess` was provided**, skip single-skill selection and run the batch assessment instead:
-
-1. Extract skill profiles (deterministic facts only):
-
-```bash
-python3 ${CLAUDE_SKILL_DIR}/scripts/assess_skills.py --json
-```
-
-This discovers all project skills and extracts metadata: allowed tools, script counts, capability flags, existing eval status, and a body excerpt from each SKILL.md. If the output is an empty JSON array, report that no skills were found and stop.
-
-2. Read the assessment prompt:
-
-```
-Read ${CLAUDE_SKILL_DIR}/prompts/assess-skills.md
-```
-
-3. Using the criteria from the assessment prompt, evaluate each skill profile from the JSON output. For skills already marked with `recommendation: "EXISTS"`, carry that verdict through. For all others, assign RECOMMENDED, OPTIONAL, or SKIP with a one-line rationale based on what the skill actually does — not on tool counts or keyword matches alone.
-
-4. Display the grouped assessment report to the user. Offer to run `/eval-analyze --skill <name>` on any of the RECOMMENDED skills. Stop here — do not proceed to Step 2.
 
 **If `--skill` was provided**, locate its SKILL.md:
 
--- a/skills/eval-analyze/SKILL.md
+++ b/skills/eval-analyze/SKILL.md
@@ -42,12 +42,6 @@ If `--config` provided, use that path. Otherwise run `python3 ${CLAUDE_SKILL_DIR
 
 Set the resolved config path as `<config>` for all subsequent steps. Set `<eval_md_path>` to the same directory as `<config>`, with filename `eval.md`.
 
-**Modes**: 
-- `--skill my-skill` → skill-based eval (`execution.skill`, case/batch mode)
-- `--prompt examples/openshift-agentic-docs.md` → prompt-based eval (`execution.prompt`, case mode, tests agent capabilities)
-- `--assess` → batch assessment of all skills (no eval.yaml generated)
-- See `examples/` for domain-specific analysis prompt recipes
-
 ## Batch Assessment
 
 If `--assess` was provided (even with `--skill`, which is ignored in this mode):
--- a/skills/eval-analyze/SKILL.md
+++ b/skills/eval-analyze/SKILL.md
@@ -81,16 +81,7 @@ python3 ${CLAUDE_SKILL_DIR}/scripts/find_skills.py
 
 This reads `.claude-plugin/plugin.json` for custom skill paths, falls back to `.claude/skills/` and `skills/`, and excludes eval harness skills. If only one skill is found, use it automatically. If none are found, tell the user — they may need to check their skill directory paths or create a skill first.
 
-If multiple skills are found, ask the user which strategy to use via AskUserQuestion:
-
-- **"Analyze a specific skill"** — list the skills and let the user pick one (current behavior)
-- **"Assess all skills — recommend which ones need evals"** — run the batch assessment:
-
-```bash
-python3 ${CLAUDE_SKILL_DIR}/scripts/assess_skills.py
-```
-
-Display the assessment report and offer to run full `/eval-analyze` on any of the recommended skills. If the user picks a skill to analyze, continue to Step 2. Otherwise stop here.
+If multiple skills are found, list them and ask the user which one to analyze via AskUserQuestion. For batch assessment, use `--assess` explicitly.
 
 **If `--update` and eval.yaml already has a `skill` field**: use that skill. If `--skill` is also provided and differs, ask the user which they mean — don't silently overwrite.
 
--- a/skills/eval-analyze/SKILL.md
+++ b/skills/eval-analyze/SKILL.md
@@ -59,13 +59,27 @@ Set the resolved config path as `<config>` for all subsequent steps. Set `<eval_
 
 ### Step 2: Find the Target Skill
 
+**If both `--assess` and `--skill` are provided**, warn the user that `--assess` performs batch assessment of all skills and `--skill` is ignored for this mode. Proceed with `--assess`.
+
 **If `--assess` was provided**, skip single-skill selection and run the batch assessment instead:
 
+1. Extract skill profiles (deterministic facts only):
+
 ```bash
-python3 ${CLAUDE_SKILL_DIR}/scripts/assess_skills.py
+python3 ${CLAUDE_SKILL_DIR}/scripts/assess_skills.py --json
 ```
 
-This reads all SKILL.md files in the project, extracts eval-relevant metadata (allowed tools, line count, complexity signals), and scores each skill to recommend whether evals would add value. Display the report to the user and offer to run full `/eval-analyze` on any of the recommended skills. Stop here — do not proceed to Step 2.
+This discovers all project skills and extracts metadata: allowed tools, script counts, capability flags, existing eval status, and a body excerpt from each SKILL.md. If the output is an empty JSON array, report that no skills were found and stop.
+
+2. Read the assessment prompt:
+
+```
+Read ${CLAUDE_SKILL_DIR}/prompts/assess-skills.md
+```
+
+3. Using the criteria from the assessment prompt, evaluate each skill profile from the JSON output. For skills already marked with `recommendation: "EXISTS"`, carry that verdict through. For all others, assign RECOMMENDED, OPTIONAL, or SKIP with a one-line rationale based on what the skill actually does — not on tool counts or keyword matches alone.
+
+4. Display the grouped assessment report to the user. Offer to run `/eval-analyze --skill <name>` on any of the RECOMMENDED skills. Stop here — do not proceed to Step 2.
 
 **If `--skill` was provided**, locate its SKILL.md:
 
--- a/skills/eval-analyze/scripts/assess_skills.py
+++ b/skills/eval-analyze/scripts/assess_skills.py
@@ -115,7 +115,9 @@ def assess_all():
                 "uses_orchestration": bool(tools & ORCHESTRATION_TOOLS),
                 "script_count": _count_scripts(path),
                 "has_existing_eval": has_eval,
-                "skill_body_excerpt": body.strip()[:BODY_EXCERPT_LIMIT],
+                "skill_body_excerpt": "<<<EXCERPT>>>"
+                + body.strip()[:BODY_EXCERPT_LIMIT]
+                + "<<<END_EXCERPT>>>",
             }
 
             if has_eval:
--- a/skills/eval-analyze/scripts/assess_skills.py
+++ b/skills/eval-analyze/scripts/assess_skills.py
@@ -1,0 +1,224 @@
+#!/usr/bin/env python3
+"""Assess which project skills are worth creating evals for.
+
+Reads all SKILL.md files, extracts eval-relevant metadata (tools, complexity,
+output type), and scores each skill to recommend whether evals would add value.
+
+Usage:
+    python3 ${CLAUDE_SKILL_DIR}/scripts/assess_skills.py [--json]
+"""
+
+import agent_eval._bootstrap  # noqa: F401 — auto-activate venv
+
+import argparse
+import json
+import sys
+from pathlib import Path
+
+import yaml
+
+# Import skill discovery from sibling module
+sys.path.insert(0, str(Path(__file__).parent))
+from find_skills import list_skills
+
+COMPLEXITY_KEYWORDS = {
+    "diagnose", "classify", "threat", "score",
+    "assessment", "investigate", "severity", "root cause",
+    "stride", "mitre",
+}
+
+FILE_TOOLS = {"Write", "Edit", "NotebookEdit"}
+EXEC_TOOLS = {"Bash"}
+AGENT_TOOLS = {"Agent"}
+
+
+def _parse_frontmatter(content):
+    """Extract frontmatter dict from SKILL.md content."""
+    if not content.startswith("---"):
+        return {}
+    parts = content.split("---", 2)
+    if len(parts) < 3:
+        return {}
+    try:
+        return yaml.safe_load(parts[1]) or {}
+    except yaml.YAMLError:
+        return {}
+
+
+def _parse_tools(fm):
+    """Extract allowed-tools list from frontmatter."""
+    tools_raw = fm.get("allowed-tools", "")
+    if isinstance(tools_raw, list):
+        return set(tools_raw)
+    return {t.strip() for t in str(tools_raw).split(",") if t.strip()}
+
+
+def _find_complexity_signals(body):
+    """Find complexity keywords in the skill body (case-insensitive)."""
+    body_lower = body.lower()
+    return sorted(kw for kw in COMPLEXITY_KEYWORDS if kw in body_lower)
+
+
+def _is_thin_wrapper(fm, line_count):
+    """Detect skills that are thin wrappers around a single command."""
+    desc = str(fm.get("description", "")).lower()
+    thin_signals = ["simple", "greeting", "label a ", "list "]
+    return line_count < 50 and any(s in desc for s in thin_signals)
+
+
+def _has_existing_eval(skill_path):
+    """Check if an eval.yaml already exists for this skill."""
+    skill_dir = Path(skill_path).parent
+    plugin_dir = skill_dir.parent.parent
+    skill_name = skill_dir.name
+
+    candidates = [
+        plugin_dir / "evals" / skill_name / "eval.yaml",
+        plugin_dir / "evals" / f"{skill_name}.yaml",
+        skill_dir / "eval.yaml",
+    ]
+    return any(c.exists() for c in candidates)
+
+
+def score_skill(meta):
+    """Compute eval-worthiness score from skill metadata."""
+    if meta["is_thin_wrapper"]:
+        return 0, [f"{meta['line_count']} lines", "thin wrapper"]
+
+    score = 0
+    reasons = []
+
+    if meta["produces_files"]:
+        score += 1
+        reasons.append("files")
+    if meta["uses_agents"]:
+        score += 1
+        reasons.append("agents")
+    if meta["line_count"] > 200:
+        score += 1
+    if meta["line_count"] > 400:
+        score += 1
+    reasons.append(f"{meta['line_count']} lines")
+
+    signals = meta["complexity_signals"]
+    signal_score = min(len(signals), 3)
+    score += signal_score
+    if signals:
+        reasons.append("/".join(signals[:3]))
+
+    return score, reasons
+
+
+def recommend(score, has_eval):
+    """Map score to recommendation."""
+    if has_eval:
+        return "EXISTS"
+    if score >= 4:
+        return "RECOMMENDED"
+    if score >= 2:
+        return "OPTIONAL"
+    return "SKIP"
+
+
+def assess_all():
+    """Assess all project skills and return structured results."""
+    skills = list_skills()
+    results = []
+
+    for skill in skills:
+        path = skill["path"]
+        try:
+            content = Path(path).read_text()
+        except OSError:
+            continue
+
+        fm = _parse_frontmatter(content)
+        tools = _parse_tools(fm)
+
+        parts = content.split("---", 2)
+        body = parts[2] if len(parts) >= 3 else content
+        line_count = len(content.splitlines())
+
+        meta = {
+            "name": skill["name"],
+            "path": skill["path"],
+            "description": skill.get("description", ""),
+            "line_count": line_count,
+            "allowed_tools": sorted(tools),
+            "produces_files": bool(tools & FILE_TOOLS),
+            "uses_bash": bool(tools & EXEC_TOOLS),
+            "uses_agents": bool(tools & AGENT_TOOLS),
+            "has_existing_eval": _has_existing_eval(path),
+            "complexity_signals": _find_complexity_signals(body),
+            "is_thin_wrapper": _is_thin_wrapper(fm, line_count),
+        }
+
+        score, reasons = score_skill(meta)
+        meta["eval_score"] = score
+        meta["score_reasons"] = reasons
+        meta["recommendation"] = recommend(score, meta["has_existing_eval"])
+
+        results.append(meta)
+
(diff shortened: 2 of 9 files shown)


kasturinarra:
@coderabbitai review

kasturinarra:
@coderabbitai review

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def _is_thin_wrapper(fm, line_count):
+    """Detect skills that are thin wrappers around a single command."""
+    desc = str(fm.get("description", "")).lower()
+    thin_signals = ["simple", "greeting", "label a ", "list "]
+    return line_count < 50 and any(s in desc for s in thin_signals)
+
+
+def _has_existing_eval(skill_path):
```
astefanutti: **`_has_existing_eval` misses the root layout and keys on the dir name, not the `skill:` field — wrong EXISTS.** It checks only `eval/<dir>/eval.yaml`, `eval/<dir>.yaml`, and a non-standard `<skill_dir>/eval.yaml` — never the root `eval.yaml` (the single-eval default, spec 005) — and matches on the SKILL.md **directory name**, while the canonical `agent_eval.config.discover_configs()` keys identity on the `skill:` field *inside* the YAML (`eval_name = raw["skill"]`). Concrete failure: `skills/foo/SKILL.md` + a root `eval.yaml` with `skill: foo` → returns `False` → the skill is recommended for an eval that already exists. It also hardcodes `Path.cwd()` (line 73), so out-of-tree plugin/marketplace skills never match, and a dir-name collision across plugins yields a false EXISTS.

**Fix:** compute once before the loop and match on the unique identifier:
```python
from agent_eval.config import discover_configs
_eval_names = {r.eval_name for r in discover_configs(Path.cwd())}
# then: has_eval = skill["dir_name"] in _eval_names
```
(Use `dir_name`, not the display `name` — see the name/dir_name comment.)

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def assess_all():
+    """Assess all project skills and return structured results."""
+    skills = list_skills()
+    results = []
+
+    for skill in skills:
+        path = skill["path"]
+        try:
```
astefanutti, line 139: **One malformed SKILL.md aborts the *entire* assessment.** The `try/except` here wraps only `read_text()` and catches only `OSError`; everything after (`_parse_frontmatter`, `_parse_tools`, `score_skill`, `sorted(tools)`) runs unguarded inside the loop, and `assess_all`/`main` have no outer guard. So any per-skill parse/type error propagates out and kills the whole run instead of skipping that skill. Two concrete triggers below (non-dict frontmatter, unhashable tools). Wrap the per-skill body in `try/except Exception: continue` (with a warning), matching how `find_skills.list_skills` tolerates bad files.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+
+def _parse_frontmatter(content):
+    """Extract frontmatter dict from SKILL.md content."""
+    if not content.startswith("---"):
+        return {}
+    parts = content.split("---", 2)
+    if len(parts) < 3:
+        return {}
+    try:
+        return yaml.safe_load(parts[1]) or {}
```
astefanutti: **Non-dict frontmatter → `AttributeError` (aborts the run, per the loop-guard comment).** `yaml.safe_load(parts[1]) or {}` only converts a falsy `None`; a YAML **sequence** or bare **scalar** frontmatter parses to a truthy `list`/`str`, which survives `or {}`. Then `_parse_tools`→`fm.get("allowed-tools", …)` (line 50) raises `AttributeError: 'list' object has no attribute 'get'`. Guard with `data = yaml.safe_load(...); return data if isinstance(data, dict) else {}`.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+
+def assess_all():
+    """Assess all project skills and return structured results."""
+    skills = list_skills()
+    results = []
+
+    for skill in skills:
+        path = skill["path"]
+        try:
+            content = Path(path).read_text()
```
astefanutti: **`UnicodeDecodeError` is not caught by `except OSError` → one non-UTF-8 SKILL.md aborts the run.** `read_text()` defaults to UTF-8 strict, and `UnicodeDecodeError` is a `ValueError`, not an `OSError`, so it escapes the handler on line 132. Note `list_skills()` already read the same file under `except Exception` and survived, so the file is known-listable but fatal to this re-read. Use `read_text(encoding="utf-8", errors="replace")` or widen the except.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+        return yaml.safe_load(parts[1]) or {}
+    except yaml.YAMLError:
+        return {}
+
+
+def _parse_tools(fm):
+    """Extract allowed-tools list from frontmatter."""
+    tools_raw = fm.get("allowed-tools", "")
+    if isinstance(tools_raw, list):
+        return set(tools_raw)
```
astefanutti: **`set(tools_raw)` raises `TypeError` on an unhashable list item (aborts the run).** `allowed-tools` given as a YAML list containing a mapping (`- name: Read`) → `TypeError: unhashable type: 'dict'`, uncaught. Low-realism input, but same root cause as the loop-guard finding. Coerce elements to str / filter non-strings before `set(...)`.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+        return {}
+    try:
+        return yaml.safe_load(parts[1]) or {}
+    except yaml.YAMLError:
+        return {}
+
+
+def _parse_tools(fm):
+    """Extract allowed-tools list from frontmatter."""
+    tools_raw = fm.get("allowed-tools", "")
```
astefanutti: **A missing `allowed-tools` is treated as "no tools" — but in Claude Code it means UNRESTRICTED (all tools).** `fm.get("allowed-tools", "")` → empty set → `produces_files`/`uses_agents`/`uses_bash` all False, so a skill that omits the field (and can write files / spawn agents) is systematically under-scored (e.g. 250-line unrestricted file+agent skill scores 1 → SKIP). Treat absent `allowed-tools` as the unrestricted set, or infer capability from the body.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+def _find_complexity_signals(body):
+    """Find complexity keywords in the skill body (case-insensitive)."""
+    body_lower = body.lower()
+    return sorted(kw for kw in COMPLEXITY_KEYWORDS if kw in body_lower)
+
+
+def _is_thin_wrapper(fm, line_count):
+    """Detect skills that are thin wrappers around a single command."""
+    desc = str(fm.get("description", "")).lower()
+    thin_signals = ["simple", "greeting", "label a ", "list "]
```
astefanutti: **`_is_thin_wrapper` substring-matches and its early return zeroes complex short skills.** `thin_signals = ["simple","greeting","label a ","list "]` are matched as substrings (so `"list "` also hits `specialist `/`checklist `), gated only on `<50` lines, and `score_skill` returns `(0, …)` **before** any other signal is considered. A 45-line "List and triage issues, then write a report" skill (produces files, uses Agent) → thin wrapper → score 0 → SKIP. Make the signals word-boundary/description-anchored and let real signals override.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+import sys
+from pathlib import Path
+
+import yaml
+
+# Import skill discovery from sibling module
+sys.path.insert(0, str(Path(__file__).parent))
+from find_skills import list_skills
+
+COMPLEXITY_KEYWORDS = {
```
astefanutti: **`COMPLEXITY_KEYWORDS` are security/threat-modeling specific** (`threat, severity, stride, mitre, root cause`), yet they're the dominant score component (up to +3 of ~6 max). A data/refactor/codegen skill matches none → 0 complexity points → capped at OPTIONAL even when large and multi-agent, while any security skill is inflated toward RECOMMENDED. This contradicts CLAUDE.md's "generic framework, no project-specifics" principle — make the keyword set configurable (e.g. via eval.yaml) or domain-neutral.

skills/eval-analyze/SKILL.md
```diff
@@ -43,19 +44,38 @@ Set the resolved config path as `<config>` for all subsequent steps. Set `<eval_
...
+**If neither `--skill` nor `--assess` was provided**, list all project skills:
 
 ```bash
 python3 ${CLAUDE_SKILL_DIR}/scripts/find_skills.py
 ```
 
-This reads `.claude-plugin/plugin.json` for custom skill paths, falls back to `.claude/skills/` and `skills/`, and excludes eval harness skills. If only one skill is found, use it automatically. If multiple, ask the user which to analyze. If none are found, tell the user — they may need to check their skill directory paths or create a skill first.
+This reads `.claude-plugin/plugin.json` for custom skill paths, falls back to `.claude/skills/` and `skills/`, and excludes eval harness skills. If only one skill is found, use it automatically. If none are found, tell the user — they may need to check their skill directory paths or create a skill first.
+
+If multiple skills are found, ask the user which strategy to use via AskUserQuestion:
```
astefanutti: **Behavior change for the auto-bootstrap path: a mandatory `AskUserQuestion` with no auto-answerer.** The old flow was "multiple skills → ask which to analyze" (one prompt); this inserts a mandatory strategy `AskUserQuestion` first. `/eval-analyze` is auto-invoked by `/eval-run` when `eval.yaml` is missing (and transitively by `/eval-optimize`), and the 3-tier AskUserQuestion answerer only applies to the *skill under test at run time*, not eval-analyze's own prompts. So in automated/CI paths this extra prompt stalls or auto-picks — a regression from the prior deterministic multi-skill→pick flow. Consider skipping the strategy prompt in the non-interactive/auto-invoked path.

skills/eval-analyze/SKILL.md
```diff
@@ -43,19 +44,38 @@ Set the resolved config path as `<config>` for all subsequent steps. Set `<eval_
 
 ## Step 1: Find the Target Skill
 
-If `--skill` was provided, locate its SKILL.md:
+**If `--assess` was provided**, skip single-skill selection and run the batch assessment instead:
```
astefanutti: **`--assess` + `--skill` silently drops `--skill`, and the two assess entry points contradict each other.** `--assess` is checked first and says "Stop here — do not proceed to Step 2" (line 53), so `/eval-analyze --skill foo --assess` runs whole-project assessment and ignores `foo` with no message. Meanwhile the multi-skill branch runs the *same* assessment then says "If the user picks a skill to analyze, continue to Step 2" (line 78) — so branch A's own "offer to run full /eval-analyze" is a dead end while branch B continues. An agent will execute the two paths inconsistently. Reconcile the continuation rule and document `--assess`/`--skill` precedence.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+        description=__doc__,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    parser.add_argument("--json", action="store_true",
+                        help="Output raw JSON instead of formatted report")
+    args = parser.parse_args()
+
+    results = assess_all()
+
+    if not results:
```
astefanutti: **`--json` on an empty project prints text to stderr and exits 1 instead of `[]`.** The `if not results` guard runs before the `--json` branch, so a consumer piping `assess_skills.py --json` into a parser gets empty stdout + exit 1 and can't distinguish "no skills" from a crash — contrary to the PR's own test plan ("`--json` produces valid JSON"). Emit `[]` (and exit 0) in JSON mode when there are no results.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+    """Extract allowed-tools list from frontmatter."""
+    tools_raw = fm.get("allowed-tools", "")
+    if isinstance(tools_raw, list):
+        return set(tools_raw)
+    return {t.strip() for t in str(tools_raw).split(",") if t.strip()}
+
+
+def _find_complexity_signals(body):
+    """Find complexity keywords in the skill body (case-insensitive)."""
+    body_lower = body.lower()
```
astefanutti: **Substring keyword matching over-matches.** `"score" in body_lower` also matches "underscore"/"scoring"; "assessment" matches "reassessment". A body merely saying "we score the results" gains spurious complexity credit (up to +3), inflating SKIP→OPTIONAL/RECOMMENDED. Use word-boundary matching (e.g. a compiled `\b(...)\b` regex).

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+
+COMPLEXITY_KEYWORDS = {
+    "diagnose", "classify", "threat", "score",
+    "assessment", "investigate", "severity", "root cause",
+    "stride", "mitre",
+}
+
+FILE_TOOLS = {"Write", "Edit", "NotebookEdit"}
+EXEC_TOOLS = {"Bash"}
+AGENT_TOOLS = {"Agent"}
```
astefanutti, line 28: **The `Skill` tool (sub-skill orchestration) is never scored.** Only `Agent` counts for orchestration, but composition via the `Skill` tool — used by eval-run/eval-optimize themselves — is one of the strongest "worth an eval" signals. A 180-line `release-pipeline` skill using `Read, Bash, Skill` scores 0 → SKIP. Add `Skill` (and reconsider whether `Bash` should score — see the `uses_bash` comment).

astefanutti (commented):
## Follow-up deep review of `--assess`

9-angle finder pass + per-finding verification + a gap sweep. **15 findings inline below**, ranked most-severe first (correctness → heuristic soundness → cleanup). Overall this is a clean, useful addition — the blocker is #1 (wrong EXISTS in the common root layout); #2–#5 are trivial hardening; the scoring heuristics (#6–#12) will misrecommend on non-security projects.

**Highlights:**
- `_has_existing_eval` misses the root `eval.yaml` layout and keys on dir name instead of the `skill:` field → reuse `agent_eval.config.discover_configs()`.
- Three "one bad SKILL.md aborts the whole run" paths (non-dict frontmatter, non-UTF-8 read, unhashable tools) — the per-skill loop needs a guard.
- Scoring bias: missing `allowed-tools` = unrestricted (not no-tools); security-only complexity keywords; `Skill` tool unscored; line-count proxy; substring over-matching.
- SKILL.md flow: a new mandatory `AskUserQuestion` regresses the auto-bootstrap path; `--assess`/`--skill` precedence and "stop vs continue" are contradictory.

**Also verified but cut for brevity (lower severity):** frontmatter parsing now duplicated a 4th time in the repo (extract one helper); each SKILL.md is read + YAML-parsed twice per run and the `---` split done twice; `score_reasons` don't reflect the line-count score bumps; category strings duplicated in 4 places; a UTF-8 BOM makes `startswith("---")` false; the CWD=project-root assumption is undocumented; `--assess` isn't persisted to `state.py`.

**Refuted (not a bug):** `AGENT_TOOLS = {"Agent"}` is correct — this repo's skills declare `Agent` (not `Task`) in `allowed-tools`.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+    score = 0
+    reasons = []
+
+    if meta["produces_files"]:
+        score += 1
+        reasons.append("files")
+    if meta["uses_agents"]:
+        score += 1
+        reasons.append("agents")
+    if meta["line_count"] > 200:
```
astefanutti: **Line count over the whole SKILL.md is a weak complexity proxy.** `line_count` counts frontmatter + prose; +1 at >200 and +2 at >400. A 420-line mostly-prose skill wrapping one command gets +2 → OPTIONAL/RECOMMENDED, while a dense 90-line multi-tool skill scores low. The real logic lives in `scripts/`, which this never inspects — consider counting referenced scripts or tool diversity instead of doc length.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,224 @@
...
+        line_count = len(content.splitlines())
+
+        meta = {
+            "name": skill["name"],
+            "path": skill["path"],
+            "description": skill.get("description", ""),
+            "line_count": line_count,
+            "allowed_tools": sorted(tools),
+            "produces_files": bool(tools & FILE_TOOLS),
+            "uses_bash": bool(tools & EXEC_TOOLS),
```
astefanutti: **`uses_bash` is computed but never used by `score_skill`.** `EXEC_TOOLS`/`uses_bash` feed only the `--json` output; Bash-side-effect skills (deploy/codegen — exactly what evals should guard) get no score credit. Either wire it into scoring or drop the field + constant. (Related: `meta["name"]` on line 143 uses the frontmatter **display name** while `_has_existing_eval` and the sort tiebreaker on line 163 key on the **dir name** / a possibly-non-unique display name — carry `skill["dir_name"]` through and use it for matching, sorting, and the `--skill` follow-up.)

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Done.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/SKILL.md
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
kasturinarra: Addressed.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,267 @@
...
+            count += 1
+    return count
+
+
+def _build_eval_names():
+    """Build the set of eval names from discover_configs (skill: field)."""
+    return {r.eval_name for r in discover_configs(Path.cwd())}
+
+
+def score_skill(meta):
```
astefanutti: This function is the crux of the architectural point in the review summary. `score_skill` is making a *judgment* — "is this worth evaluating?" — out of proxies (tool booleans, tool count, script count, capped complexity-keyword hits, fixed cutoffs). It's the part I'd move to a single batched LLM call (with these same features passed as inputs) rather than continue tuning: the model can weigh "non-deterministic, quality-sensitive output" directly instead of via correlates. The deterministic facts feeding it — `has_existing_eval` (from `discover_configs`), the tool profile, `script_count` — are exactly the right things to compute in Python and hand to the judge.

astefanutti (commented):
## A step-back question on the approach

Before iterating further on the scorer, though, I want to raise an architectural question, because I think it dissolves a whole class of the findings rather than patching them one by one.

### The scoring is a *judgment* task implemented as *deterministic proxies*

`assess_skills.py` answers "would evals add value for this skill?" using tool counts, script counts, a keyword regex, and threshold cutoffs. Those aren't the thing we care about — they're weak stand-ins for it. The review history is itself the tell: line-count → tool-diversity → which now over-rewards skills that simply omit `allowed-tools` (a skill with no `allowed-tools` is treated as all-6-tools and auto-scores RECOMMENDED, while a skill that declares `allowed-tools: Read` scores 0/SKIP). Each proxy fix surfaces the next proxy's failure mode. That's the classic sign the task is a poor fit for a fixed heuristic.

It's also inconsistent with the rest of `eval-analyze`, which follows the harness philosophy (CLAUDE.md: *schema-driven, observe don't assume, the LLM interprets, scripts just move files*). Step 3 already spawns an Explore agent to judge the skill; `--assess` is the one place that hardcodes a judgment in Python.

Concretely, the heuristic mis-fires both ways:
- **False SKIP:** a 40-line skill that writes one carefully-worded customer email — 1 tool, no scripts, short → SKIP. But that non-deterministic, quality-sensitive output is exactly what a linter can't catch and an eval should. Highest-value target, scored lowest.
- **False RECOMMENDED:** a 600-line skill that scaffolds boilerplate with deterministic file output — many tools/scripts → RECOMMENDED, even though a schema/lint check already covers it.

### What should stay deterministic vs. move to an LLM/prompt

The split isn't "all or nothing" — it's **facts stay in Python, judgment moves to the model**:

**Keep deterministic (Python) — objective facts that must be exact and reproducible:**
- **Skill discovery** — `find_skills.list_skills()` (dir globbing, plugin.json). Pure file I/O.
- **Existing-eval detection** — `discover_configs()` matched on the `skill:` field. This is the strongest keep-in-code item: whether an `eval.yaml` exists is a fact, and an LLM must never guess it. The EXISTS category should always come from here.
- **Raw feature extraction as *inputs*** — declared `allowed-tools`, script/file counts, frontmatter fields. Objective observations — but pass them to the judge, don't threshold them into the verdict.
- **Output plumbing** — the `--json` shape and exit codes.

**Move to an LLM/prompt/skill — the judgment:**
- **The RECOMMENDED / OPTIONAL / SKIP verdict** — "is this skill's output complex / non-deterministic / quality-sensitive enough that evals catch regressions a linter wouldn't?"
- **Complexity assessment** — instead of a maintained keyword regex, let the model read the body and judge what the skill actually does.
- **Thin-wrapper detection** — "is this just a shim around one command?" is a judgment, not a description-substring test.
- **The per-skill rationale.**

### But keep it cheap — it's a pre-filter

`--assess` exists to decide *which* skills are worth the expensive per-skill analysis, so it shouldn't do that analysis for all of them. The cheap-and-accurate shape is **one batched LLM call**: feed every skill's name + description + the deterministic features (has-eval, tool profile, script count) in a single prompt, get back the ranked recommendations with one-line reasons. ~1 call total, not N — and it reasons about actual eval-worthiness instead of a formula. The most `eval-analyze`-native version is to drop the Python scorer entirely and make `--assess` a SKILL.md instruction, with `find_skills.py` + `discover_configs()` as the only (deterministic) scripts.

This would also make the missing-`allowed-tools` inflation, the keyword-list maintenance, and the threshold tuning all go away — the model judges the thing directly.

### If you'd rather keep the deterministic scorer for now

Totally reasonable as a v1 — in that case these residuals are worth a quick fix (happy to detail any):
1. Missing `allowed-tools` → all-tools → auto-RECOMMENDED (the interaction above). Treat *absent* `allowed-tools` as neutral for the tool-diversity score, not as "uses everything."
2. Explicit empty `allowed-tools:` (null) → phantom `{'None'}` tool — treat like the missing case.
3. Step 1 of SKILL.md still says assess extracts "line count" — that's no longer a scoring signal.
4. `--assess` + `--skill` together silently ignores `--skill` (no precedence guidance).

Not a merge blocker either way — mainly flagging that the redesign is likely less total effort than continuing to tune the heuristic, and more accurate.

astefanutti (commented):
@kasturinarra thanks for the update, I've left a comment on the architecture / design, suggesting splitting facts with Python and moving the judgement part to the LLM / skill that's best for that.

Let me know what you think.

kasturinarra:
> @kasturinarra thanks for the update, I've left a comment on the architecture / design, suggesting splitting facts with Python and moving the judgement part to the LLM / skill that's best for that.
> 
> Let me know what you think.

@astefanutti Thanks for the thorough analysis — I agree with the architectural split. The heuristic approach was a reasonable starting point, but you're right that each proxy fix just surfaces the next failure mode, and it's inconsistent with the rest of eval-analyze's design.

 I'll rework --assess to:
  - Keep in Python: skill discovery, existing-eval detection (discover_configs), raw feature extraction (tools, script counts, frontmatter fields), and output plumbing (--json shape, exit codes)
  - Move to LLM: the RECOMMENDED/OPTIONAL/SKIP verdict, complexity assessment, thin-wrapper detection, and per-skill rationale — via a single batched call with all skills' metadata, keeping  it cheap as a pre-filter

would you like me to add a new PR or update the same with the above design ?

astefanutti:
@kasturinarra thanks, that sounds good.

I think you can update this PR, but if you have a preference for opening a new one that works too.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,183 @@
...
+def assess_all():
+    """Extract metadata profiles for all project skills."""
+    skills = list_skills()
+    eval_names = _build_eval_names()
+    results = []
+
+    for skill in skills:
+        path = skill["path"]
+        try:
+            content = Path(path).read_text(errors="replace")
```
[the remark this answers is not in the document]
kasturinarra: Skipping this finding — not a valid issue in context.

  The path values in assess_all() come from list_skills() in find_skills.py, which constructs them via glob(f"{skills_dir}/*/SKILL.md") — a filesystem glob over known skill directories. The
  skill directories themselves are already validated:

  1. plugin.json path — passes through _resolve_under_cwd() (line 31-36), which resolves the path and rejects anything outside Path.cwd() via relative_to(). Symlinks are resolved by
  .resolve().
  2. marketplace.json path — also goes through _resolve_under_cwd() (line 85).
  3. Default paths — hardcoded [".claude/skills", "skills"], filtered to is_dir().

  The glob pattern {skills_dir}/*/SKILL.md only matches direct children named SKILL.md — it can't traverse upward. A symlink inside a skill dir pointing outside CWD is already handled by the
  upstream _resolve_under_cwd guard on the parent directory.
  
  Adding a second path validation in assess_skills.py would duplicate the boundary already enforced in find_skills.py without covering any new attack surface.

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,183 @@
...
+    """Print a human-readable listing of skill profiles."""
+    print("Skill Profiles")
+    print("=" * 65)
+
+    for r in results:
+        status = "[EXISTS]" if r.get("recommendation") == "EXISTS" else ""
+        tools_str = ", ".join(r["allowed_tools"])
+        print(f"\n  {r['name']} {status}")
+        print(f"    {r['description']}")
+        print(f"    tools: {tools_str}")
```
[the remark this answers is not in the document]
kasturinarra: Skipping — not a valid issue in practice.

  The values printed here (name, description, allowed_tools) all come from YAML frontmatter in SKILL.md files that the user controls within their own project. These are local files on the
  user's own machine, not external/user-submitted input. The same values are already printed unescaped by find_skills.py (list_skills → line 204: print(f"SKILL: {s['name']:<30} 
  {s['description']}")), and by the SKILL.md's own conversation output.

  This is a CLI tool running in the user's terminal on their own project files — the threat model of a malicious SKILL.md injecting escape sequences into the terminal of the person who put
  that file there doesn't hold. Sanitizing here would be inconsistent with the rest of the codebase and would add complexity for a non-existent attack surface.

skills/eval-analyze/SKILL.md
```diff
@@ -43,19 +44,43 @@ Set the resolved config path as `<config>` for all subsequent steps. Set `<eval_
...
+
+This discovers all project skills and extracts metadata: allowed tools, script counts, capability flags, existing eval status, and a body excerpt from each SKILL.md. If the output is an empty JSON array, report that no skills were found and stop.
+
+2. Read the assessment prompt:
+
+```
+Read ${CLAUDE_SKILL_DIR}/prompts/assess-skills.md
+```
+
+3. Using the criteria from the assessment prompt, evaluate each skill profile from the JSON output. For skills already marked with `recommendation: "EXISTS"`, carry that verdict through. For all others, assign RECOMMENDED, OPTIONAL, or SKIP with a one-line rationale based on what the skill actually does — not on tool counts or keyword matches alone.
```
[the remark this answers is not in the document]
kasturinarra: Skipping — already addressed. The untrusted-data boundary for skill_body_excerpt was added in the previous fix to assess-skills.md (line 11): the prompt already instructs the LLM to treat
  it as untrusted data, ignore operational directives/instructions/role assignments, and use it only to assess what the skill does. The script also wraps excerpts in
  <<<EXCERPT>>>...<<<END_EXCERPT>>> delimiters. Adding the same guidance a second time in SKILL.md would be redundant — the SKILL.md already says "Using the criteria from the assessment
  prompt" (line 65), which defers to the prompt file where the boundary is defined.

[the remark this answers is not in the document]
kasturinarra: pushed now.

kasturinarra:
@coderabbitai review

kasturinarra:
@astefanutti could you give a review when ever you get sometime ? thanks !!

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,185 @@
...
+    for skill in skills:
+        path = skill["path"]
+        try:
+            content = Path(path).read_text(errors="replace")
+
+            fm = _parse_frontmatter(content)
+            tools = _parse_tools(fm)
+            body = _extract_body(content)
+
+            has_eval = skill["dir_name"] in eval_names
```
astefanutti: **EXISTS detection likely misses skills that already have evals.** `eval_names` come from `discover_configs`, where `eval_name` is derived first from the config's `skill:` field (config.py:1090) — the invocation/frontmatter name (e.g. `rfe.create`, as used in `references/eval-yaml-template.md:46`). But this matches it against `dir_name` (the directory, e.g. `rfe-creator`). For any namespaced/plugin skill these differ, so `has_eval` is `False`, `recommendation: "EXISTS"` is never set, and the LLM re-assesses (and may recommend building) an eval that already exists.

Matching the frontmatter name too is more robust:
```python
has_eval = skill["dir_name"] in eval_names or skill["name"] in eval_names
```

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,185 @@
...
+    except yaml.YAMLError:
+        return {}
+
+
+def _parse_tools(fm):
+    """Extract allowed-tools list from frontmatter.
+
+    Missing or null allowed-tools means unrestricted (all tools) in Claude Code.
+    """
+    tools_raw = fm.get("allowed-tools")
```
astefanutti: **Two issues in `_parse_tools`, both feed wrong signals to the judge:**

1. **Scoped tool specifiers aren't normalized.** Claude Code allows `allowed-tools: Read, Bash(git:*), Edit(src/**)`. Those are kept verbatim, so `tools & EXEC_TOOLS` (`{"Bash"}`) and `tools & FILE_TOOLS` miss them → `uses_bash`/`produces_files` come out `False` for any skill using scoped-permission syntax. Strip the `(...)` suffix, e.g. `t.split("(", 1)[0].strip()`.

2. **`ALL_TOOLS` is a lossy sentinel for "unrestricted."** When `allowed-tools` is missing/null/empty-string you return `ALL_TOOLS` (just the 6 capability tools), which then lands in `allowed_tools: sorted(tools)` (line 111). The prompt tells the LLM that field is "the full tool surface," but it omits Read/Grep/Glob/etc. and reads as an explicit 6-tool declaration. Relatedly, `allowed-tools: []` returns an empty set (all flags `False`) while `allowed-tools: ""` falls through to `ALL_TOOLS` (all flags `True`) — same intent, opposite profile. A distinct `unrestricted` marker (or boolean) would avoid conflating "declared these 6" with "declared nothing."

skills/eval-analyze/scripts/assess_skills.py
```diff
@@ -0,0 +1,185 @@
...
+                "path": skill["path"],
+                "description": skill.get("description", ""),
+                "allowed_tools": sorted(tools),
+                "produces_files": bool(tools & FILE_TOOLS),
+                "uses_bash": bool(tools & EXEC_TOOLS),
+                "uses_agents": bool(tools & AGENT_TOOLS),
+                "uses_orchestration": bool(tools & ORCHESTRATION_TOOLS),
+                "script_count": _count_scripts(path),
+                "has_existing_eval": has_eval,
+                "skill_body_excerpt": "<<<EXCERPT>>>"
```
astefanutti, line 163: **The `<<<END_EXCERPT>>>` fence is spoofable by the untrusted content it's meant to isolate.** The body is the first 500 raw chars of a SKILL.md — a hostile skill can put `<<<END_EXCERPT>>>` at the top followed by injected instructions, breaking out of the fence the `assess-skills.md` prompt relies on. The "treat as untrusted data" instruction is the real defense; if you keep the delimiter as defense-in-depth, strip the closing marker from `body` first, e.g. `body.replace("<<<END_EXCERPT>>>", "")`.

skills/eval-analyze/scripts/find_skills.py
```diff
@@ -36,6 +36,15 @@ def _resolve_under_cwd(raw, base):
     return candidate
 
 
+def _is_under_cwd(path):
```
astefanutti: **This guard rejects legitimately symlinked skill directories, not just traversal escapes.** `_is_under_cwd` resolves symlinks before the containment check, so a valid monorepo/plugin-dev layout like `skills/deploy -> ../../shared-skills/deploy` now fails: `find_skill` returns `None` (the exact-path branch at line 148 skips silently — no WARNING, unlike the glob branches) and `list_skills` omits it, so the skill can no longer be analyzed or assessed. Is rejecting *all* out-of-CWD symlinks the intent, or only `../`-style escapes from a trusted skills root? If symlinked skill dirs should stay discoverable, this needs to allow a symlink target while still blocking traversal escapes. (Also note this duplicates the containment logic already in `_resolve_under_cwd`.)

astefanutti (commented):
Follow-up review with inline notes. The feature is well-factored (deterministic extraction in Python, judgment in the LLM prompt) and the symlink guard is a good instinct. Two of the comments below are correctness issues that feed silently-wrong data to the judge (`has_eval` mapping and scoped-tool parsing); the others are a security-hardening gap and a design question on the symlink guard.

astefanutti (commented):
@kasturinarra I've left a couple of comments.

There is also CI that fails with `Estimated 4,657 tokens exceeds skill warn limit of 4,500`.

If you can keep the skill down to 4500 tokens, maybe by moving some more logic in `assess-skills.md`.

kasturinarra:
> @kasturinarra I've left a couple of comments.
> 
> There is also CI that fails with `Estimated 4,657 tokens exceeds skill warn limit of 4,500`.
> 
> If you can keep the skill down to 4500 tokens, maybe by moving some more logic in `assess-skills.md`.

thank you, will take a look and respond soon.

skills/eval-analyze/scripts/find_skills.py
```diff
@@ -36,6 +36,32 @@ def _resolve_under_cwd(raw, base):
...
+
+    Always resolves symlinks before checking containment — a lexical
+    prefix match on an unresolved path would let a symlink escape CWD.
+    Returns the resolved Path if valid, None otherwise.
+
+    trusted_roots: resolved Paths for skill directories discovered via
+    plugin.json or marketplace.json (already validated by _resolve_under_cwd).
+    """
+    resolved = Path(path).resolve()
+    cwd = Path.cwd().resolve()
```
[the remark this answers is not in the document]
kasturinarra: Skipping this fix. Path.resolve(strict=False) — the default since Python 3.6 — handles symlink loops, nonexistent paths, and broken symlinks without raising. Confirmed on the project's
  Python 3.14: both symlink loops and nonexistent paths resolve silently (no RuntimeError, no OSError). The only scenario that could raise is a catastrophic filesystem failure (NFS timeout,
  EIO), which would also break the glob() calls and open() calls surrounding it — adding a guard here alone wouldn't save the discovery.

  The existing except Exception handlers on the open()/read() calls in find_skill() (line 186) and list_skills() (line 224) already provide warning-based continuation for file-level
  failures. Adding a redundant guard around resolve() would be dead error-handling code.

astefanutti (approved):
Thanks a lot @kasturinarra!

kasturinarra:
> Thanks a lot @kasturinarra!

Thanks to you too !!

PR #153, opened 2026-07-15
feat: pass GCP_SA_ACCESS_TOKEN to AnthropicVertex when available

danmcp:
Inside an OpenShell sandbox, the network policy restricts which binaries can reach external endpoints. Only binaries listed in the policy (e.g. claude, python3) can route through the gateway. Additionally, google-auth discovery fails because the GCP service account key file is not available inside the sandbox — the OpenShell provider injects a rotated access token (GCP_SA_ACCESS_TOKEN) instead.

Without this change, score.py must be run through Claude Code (claude -p) as a command executor just to get network access, which adds cost and indirection. With it, python3 can be added to the sandbox policy binary list and score.py calls Vertex AI directly.

When GCP_SA_ACCESS_TOKEN is set, pass it to AnthropicVertex as access_token. When absent, fall back to google-auth discovery (the existing behavior for non-sandbox environments).

## How Has This Been Tested?
I am currently patching this solution into my use of agent-eval-harness

A standalone test would look like:

```
import os
from anthropic import AnthropicVertex

access_token = os.environ.get('GCP_SA_ACCESS_TOKEN')
client = AnthropicVertex(
    project_id='test-project',
    region='us-east5',
    access_token=access_token or None,
)
assert client.access_token == 'fake-token', f"Expected 'fake-token', got {client.access_token!r}"
print('PASS: AnthropicVertex accepts access_token from env var')
```

Run with:
```
pip install anthropic
GCP_SA_ACCESS_TOKEN=fake-token python3 test_vertex_token.py
```

Unit tests added as part of this PR.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication when connecting to the Anthropic service on Google Cloud.
  * Supports optional access tokens supplied through the environment, while preserving existing behavior when no token is provided.

### DIFF
--- a/skills/eval-run/scripts/score.py
+++ b/skills/eval-run/scripts/score.py
@@ -1439,7 +1439,12 @@ def _get_anthropic_client():
     region = os.environ.get("CLOUD_ML_REGION", "us-east5")
     if project_id:
         from anthropic import AnthropicVertex
-        return AnthropicVertex(project_id=project_id, region=region)
+        access_token = os.environ.get("GCP_SA_ACCESS_TOKEN")
+        return AnthropicVertex(
+            project_id=project_id,
+            region=region,
+            access_token=access_token or None,
+        )
     api_key = os.environ.get("ANTHROPIC_API_KEY") or os.environ.get("ANTHROPIC_AUTH_TOKEN")
     if api_key:
         from anthropic import Anthropic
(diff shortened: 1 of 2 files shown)


danmcp:
@coderabbitai review

astefanutti (approved):
Thanks @danmcp!

PR #159, opened 2026-07-28
feat(mlflow): richer Harbor traces with user turns, thinking, and tool detail

GuyZivRH:
## Summary

Harbor's stream-json output often omits user text and full tool payloads that Claude Code's own transcripts capture. This PR uses Harbor's ATIF `trajectory.json` (when available) to backfill that detail in MLflow traces, so Harbor runs get the same trace richness as native runs.

- Seed the root prompt and per-turn `CHAIN` user spans from `trajectory.json` when stream-json lacks user text
- Emit extended-thinking blocks as `LLM` spans so MLflow graphs show the full agent turn sequence
- Include `Write`/`Edit` file payloads and richer ATIF tool-call arguments/observations in `TOOL` spans
- Attach tool names, text, and thinking on `AGENT` step spans for closer parity with `trajectory.json`
- Keep v1.0.3 `EvalConfig` compatibility in `log_results.py` (`resolve_skill()`/`eval_name()` fallback shims)
- All new behavior is opt-in via an optional `trajectory_path` kwarg — falls back cleanly to stream-json-only behavior when absent

## Test plan

- [x] `tests/test_trace_builder.py` — 9/9 pass, including 5 new tests (thinking spans, trajectory-seeded prompt/CHAIN spans, Write content preservation, ATIF observation enrichment, AGENT step outputs)
- [x] Full suite passes on top of current `main`
- [x] `ruff check` clean

## Summary by CodeRabbit

- **New Features**
  - MLflow traces can now use available Harbor `trajectory.json` to restore user turns and enrich tool inputs/outputs.
  - Added dedicated `thinking` spans for assistant reasoning (including backfilling from trajectory when needed).
  - Improved compact tool summaries (e.g., full Write bodies, Edit old/new, Read offset/limit).
  - Trace naming is now consistent across execution modes.

- **Bug Fixes**
  - Prefer trajectory-derived observations and long-form tool result text when it’s more complete than stream events.
  - Fixed span/step attribution so thinking belongs to the correct turn and timestamps stay within trace bounds.

### DIFF
--- a/agent_eval/mlflow/trace_builder.py
+++ b/agent_eval/mlflow/trace_builder.py
@@ -66,6 +66,7 @@ def _user_text_from_content(content):
 _MAX_TOOL_OUTPUT = 4000  # chars per tool result (stream or ATIF-enriched)
 _MAX_WRITE_CONTENT = 4000
 _MAX_EDIT_FRAGMENT = 2000
+_MAX_THINKING = 4000
 
 
 def _load_trajectory_json(trajectory_path):
@@ -210,8 +211,9 @@ def build_trace(stdout_path, run_result, run_id, experiment_id,
     traj_user_turns = _load_trajectory_user_steps(trajectory_path)
     if traj_user_turns:
         # Trajectory is the richer Harbor source of truth for user turns.
-        if not prompt:
-            prompt = "\n\n".join(t["message"] for t in traj_user_turns)
+        # Always derive prompt from it here too, so the root span's prompt
+        # attribute and the "user:" CHAIN spans agree on the same source.
+        prompt = "\n\n".join(t["message"] for t in traj_user_turns)
         user_turns_for_spans = traj_user_turns
     else:
         user_turns_for_spans = stream_user_turns
@@ -748,7 +750,7 @@ def build_trace(stdout_path, run_result, run_id, experiment_id,
         if llm_text:
             step_outputs["text"] = llm_text
         if thinkings:
-            step_outputs["thinking"] = thinkings[0][0]
+            step_outputs["thinking"] = thinkings[0][0][:_MAX_THINKING]
         step_span = make_span(
             trace_id, root_span_id,
             name=step_name,
@@ -774,7 +776,7 @@ def build_trace(stdout_path, run_result, run_id, experiment_id,
                 start_ns=t_start,
                 end_ns=t_end,
                 inputs={"model": model, "kind": "thinking"},
-                outputs={"thinking": think_text},
+                outputs={"thinking": think_text[:_MAX_THINKING]},
                 extra_attrs=({"mlflow.llm.model": json.dumps(model)}
                              if model else None),
             ))
--- a/agent_eval/mlflow/trace_builder.py
+++ b/agent_eval/mlflow/trace_builder.py
@@ -48,21 +48,71 @@ def make_span(trace_id, parent_id, name, span_type, start_ns, end_ns,
     }
 
 
+def _user_text_from_content(content):
+    """Extract plain user text from a stream-json message content field."""
+    if isinstance(content, str) and content.strip():
+        return content.strip()
+    if isinstance(content, list):
+        parts = []
+        for b in content:
+            if isinstance(b, dict) and b.get("type") == "text":
+                text = (b.get("text") or "").strip()
+                if text:
+                    parts.append(text)
+        if parts:
+            return "\n\n".join(parts)
+    return ""
+
+
+def _load_trajectory_user_steps(trajectory_path):
+    """Load user steps from a Harbor ATIF trajectory.json (if present).
+
+    Returns a list of ``{"message": str, "timestamp": str|None}``.
+    """
+    if trajectory_path is None:
+        return []
+    path = Path(trajectory_path)
+    if not path.is_file():
+        return []
+    try:
+        data = json.loads(path.read_text())
+    except (OSError, json.JSONDecodeError):
+        return []
+    if not isinstance(data, dict):
+        return []
+    steps = data.get("steps") or []
+    out = []
+    for step in steps:
+        if not isinstance(step, dict):
+            continue
+        if step.get("source") != "user":
+            continue
+        msg = step.get("message")
+        if not isinstance(msg, str) or not msg.strip():
+            continue
+        out.append({
+            "message": msg.strip(),
+            "timestamp": step.get("timestamp"),
+        })
+    return out
+
+
 def build_trace(stdout_path, run_result, run_id, experiment_id,
                 trace_name="", subagent_dir=None,
-                subagent_model=None):
+                subagent_model=None, trajectory_path=None):
     """Build a hierarchical MLflow Trace from the stream-json stdout log.
 
     Structure:
       root AGENT
-        ├── LLM (text response)
-        ├── TOOL (single sequential tool call)
-        ├── TASK "N parallel agents" (group of parallel calls)
-        │   ├── AGENT (subagent 1)
-        │   ├── AGENT (subagent 2)
-        │   └── ...
-        ├── LLM (text response)
+        ├── CHAIN user (skill invoke / instructions; from trajectory or stream)
+        ├── AGENT step
+        │   ├── LLM thinking
+        │   ├── TOOL ...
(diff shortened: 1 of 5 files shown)


agent_eval/mlflow/trace_builder.py
```diff
@@ -511,17 +653,25 @@ def _is_status_update(text):
...
             if _has_bg_agents and _is_status_update(seg_data):
                 # Merge status update into the current dispatch step
                 continue
             # Save previous step
-            if current_llm is not None or current_batches:
-                steps.append((current_llm, current_ts, current_context,
-                              current_batches))
-                current_batches = []
-                _has_bg_agents = False
+            _flush_step()
```
astefanutti, line 741: 🔴 **Thinking is attributed to the previous step, not its own turn.** `_flush_step()` runs here *before* `current_llm` is set, and it flushes `current_thinkings` into the step being closed. But within a turn Claude emits `[thinking, text, tool_use]`, so the thinking accumulated at line 668 belongs to *this* incoming `llm` segment — not the step that's closing.

Running this PR's `build_trace()` on `turn1 = text1 + Bash` / `turn2 = thinking2 + text2 + Write`:

```
STEP 'Step one: list files.'     → outputs.thinking = "Now I reason about writing the file"  # turn 2's ❌
STEP 'Step two: write the file.' → outputs.thinking = None                                   # missing ❌
```

Every turn with both thinking and text mis-groups the reasoning one step early, which contradicts the new docstring (thinking + tool + response under one AGENT step).

Suggested fix: buffer pending thinkings separately and move them into `current_thinkings` for the *new* step after the flush — e.g. `_flush_step()` (excluding pending), then `current_thinkings = pending; pending = []`.

tests/test_trace_builder.py
```diff
@@ -110,3 +110,213 @@ def test_returns_none_for_missing_file(self, tmp_path):
...
+            trajectory_path=traj,
+        )
+        write = next(s for s in trace["data"]["spans"] if s["name"] == "Write")
+        inputs = json.loads(write["attributes"]["mlflow.spanInputs"])
+        outputs = json.loads(write["attributes"]["mlflow.spanOutputs"])
+        assert 'print("Hello, World!")' in inputs["content"]
+        assert "Hello, World!" in outputs["result"]
+        assert "[metadata]" in outputs["result"]
+
+    def test_agent_step_spans_include_outputs(self, tmp_path):
```
astefanutti, line 327: 🔴 The new thinking tests all use turns with thinking but **no text** (`thinking + tool`), which is the one shape that groups correctly, and none assert **co-location** (that a thinking span and its turn's text/tool share one AGENT step). That's why the mis-grouping (see my comment on `trace_builder.py` `_flush_step`) passes CI.

Could you add a `thinking + text + tool` single-turn case asserting they land in the same step? It fails today and would lock in the fix.

agent_eval/mlflow/trace_builder.py
```diff
@@ -477,8 +597,27 @@ def _flush_batch():
         "attributes": root_attrs,
     }]
 
+    # User turns as CHAIN spans under root (visible in the MLflow graph).
+    user_cursor = trace_start
+    for idx, turn in enumerate(user_turns_for_spans):
+        ts = turn.get("timestamp")
+        start_ns = iso_to_ns(ts) if ts else user_cursor
```
astefanutti, line 656: 🟡 CHAIN `user:` spans take their start from `trajectory.json` timestamps but are never clamped to the root window `[trace_start, trace_end]` (derived from stream-json), and the root isn't widened to include them. With an ATIF clock offset I saw CHAIN spans start ~3600s *before* the root span — rendering outside/left of the parent in the MLflow timeline. The thinking spans below (764+) have the same exposure. Usually both clocks are close (same container), but a clamp would harden it.

agent_eval/mlflow/trace_builder.py
```diff
@@ -567,25 +720,65 @@ def _is_status_update(text):
...
             start_ns=step_start,
             end_ns=step_end,
-            inputs={"step": step_idx + 1},
+            inputs=step_inputs,
+            outputs=step_outputs or None,
         )
         step_span_id = step_span["span_id"]
         spans.append(step_span)
 
+        # Thinking spans (extended thinking / reasoning_content)
```
astefanutti, line 843: 🟡 These thinking spans come **only** from stream-json `thinking` blocks — ATIF `reasoning_content` is never read (it appears just in this comment and a test fixture). If a Harbor agent's stream omits thinking but `trajectory.json` carries `reasoning_content`, the thinking is silently dropped — the exact "stream is thin, backfill from trajectory" case this PR targets. Consider backfilling from `reasoning_content` when a step has no stream thinking, or a comment noting the deliberate limitation.

agent_eval/mlflow/trace_builder.py
```diff
@@ -48,22 +47,123 @@ def make_span(trace_id, parent_id, name, span_type, start_ns, end_ns,
...
+            if isinstance(b, dict) and b.get("type") == "text":
+                text = (b.get("text") or "").strip()
+                if text:
+                    parts.append(text)
+        if parts:
+            return "\n\n".join(parts)
+    return ""
+
+
+_MAX_TOOL_OUTPUT = 4000  # chars per tool result (stream or ATIF-enriched)
```
astefanutti, line 77: 🟡 `_MAX_TOOL_OUTPUT` 500 → 4000 (8×), plus full `Write` content (4000) and `Edit` fragments (2×2000) now embedded in span **inputs** (1048+), apply to **all** runs — only `trajectory_path` is Harbor-gated in `log_results.py`. For write-heavy runs this multiplies per-trace payload / MLflow UI weight. Fine if intended as a general richness bump; just flagging it's cross-cutting, not Harbor-scoped.

astefanutti (commented):
Thanks @GuyZivRH! I've left a couple of comments otherwise looks good overall.

agent_eval/mlflow/trace_builder.py
```diff
@@ -86,29 +186,36 @@ def build_trace(stdout_path, run_result, run_id, experiment_id,
...
+                prompt = text
+
+    traj_user_turns = _load_trajectory_user_steps(trajectory_path)
+    if traj_user_turns:
+        # Trajectory is the richer Harbor source of truth for user turns.
+        if not prompt:
+            prompt = "\n\n".join(t["message"] for t in traj_user_turns)
+        user_turns_for_spans = traj_user_turns
+    else:
+        user_turns_for_spans = stream_user_turns
```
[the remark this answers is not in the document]
GuyZivRH: fixed

skills/eval-mlflow/scripts/log_results.py
```diff
@@ -390,11 +408,16 @@ def main():
...
-                                         subagent_dir=sub_dir)
+                _skill = _resolve_skill(config)
+                trace_name = (f"{_skill} ({step_key})"
+                              if _skill else step_key)
+                traj = transcript.parent / "trajectory.json"
+                trace_dict = build_trace(
+                    transcript, step_rr, step_key,
+                    experiment_id, trace_name=trace_name,
+                    subagent_dir=sub_dir,
+                    trajectory_path=traj if traj.is_file() else None,
```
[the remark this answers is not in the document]
GuyZivRH: fixed

agent_eval/mlflow/trace_builder.py
GuyZivRH: fixed

tests/test_trace_builder.py
GuyZivRH: fixed

agent_eval/mlflow/trace_builder.py
GuyZivRH: fixed, good catch!

agent_eval/mlflow/trace_builder.py
GuyZivRH: fixed

agent_eval/mlflow/trace_builder.py
GuyZivRH: it is ~4KB so i will leave it as is

astefanutti (approved):
Thanks @GuyZivRH!

PR #166, opened 2026-07-29
feat(harbor): per-model usage breakdown + fix placeholder-model accounting

astefanutti:
## What
- **`harbor/results.py` + `harbor/run.py`** — thread a `per_model_usage` breakdown through Harbor parsing (extract `modelUsage` per trial/step/job) and surface it in `run_result`, bringing the Harbor runner to parity with the local runner for the report's Model Usage table.
- **`stream_capture.py`** — `_is_real_model()` filters bracketed pseudo-models (e.g. `<synthetic>`) out of `models_seen` / per-model turn+token breakdowns and subagent-model detection, while still counting their turns for dedup/totals.

## Why
Harbor jobs produced no per-model breakdown (the report's Model Usage table was local-runner-only); and placeholder pseudo-models leaked into the local per-model table as a bogus model row.

## Test plan
⚠️ **No automated tests** for these paths yet (flagged for reviewers): Harbor `modelUsage` extraction and the `<synthetic>` filter are currently untested. Manual checks:
- A Harbor run's `run_result.json` includes `per_model_usage`.
- A run containing a `<synthetic>` turn shows no bogus model row in the per-model table (turns still counted in totals).

### DIFF
--- a/agent_eval/agent/stream_capture.py
+++ b/agent_eval/agent/stream_capture.py
@@ -49,6 +49,19 @@ def inject_timestamp(line: str) -> str:
 
 # ── Usage extraction ─────────────────────────────────────────────────
 
+def _is_real_model(name):
+    """Whether ``name`` is a genuine model, not a Claude Code placeholder.
+
+    Claude Code stamps some assistant messages with a bracketed pseudo-model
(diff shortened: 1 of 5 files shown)


astefanutti:
@coderabbitai review

PR #179, opened 2026-08-11
fix: resolve bootstrap venv via realpath; sync validate_eval Jinja vars

RedTanny:
Skill scripts import agent_eval via scripts/agent_eval symlink. 
abspath kept the symlink path, so plugin_root became skills/.../scripts and .eval-venv was never found.
 Resolving via realpath fixes this.

Also sync validate_eval Jinja vars with score renderer. 
Replace phantom {{ input }} with {{ inputs }}; allow {{ reasoning }} and {{ evidence }} 
which score._render_jinja2_template always injects.

## Description
Two small harness bugs were found while running skill scripts via the designed `${CLAUDE_SKILL_DIR}/scripts/` path (import of `agent_eval` through the per-skill `scripts/agent_eval` symlink):

1. **`agent_eval/_bootstrap.py`** used `os.path.abspath(__file__)`, 
   which does not resolve that symlink. Plugin root was computed as `skills/<skill>/scripts`, so `.eval-venv` was not found and third-party deps looked missing under system `python3`.
3. **`skills/eval-analyze/scripts/validate_eval.py`** listed a phantom `{{ input }}` 
   standard variable and omitted `inputs` / `reasoning` / `evidence`, rejecting valid LLM judge templates that match `score._render_jinja2_template`.

## How Has This Been Tested?
The fix was tested locally with both a full skill run to ensure the `.eval-venv` is properly discovered and activated, and by running the unit tests using `pytest tests/ -v`, verifying that the behavior of `validate_eval` correctly allows `{{ inputs }}`, `{{ reasoning }}`, and `{{ evidence }}`. The `test_bootstrap.py` suite includes a new unit test for the symlink path validation.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup reliability when evaluation tools are accessed through symbolic links.
  * Corrected template validation to recognize the `inputs` variable.
* **New Features**
  * Added support for `reasoning` and `evidence` as standard template variables.
* **Documentation**
  * Updated validation guidance, warnings, and sample rendering data to reflect the expanded set of supported template variables.

### DIFF
--- a/agent_eval/_bootstrap.py
+++ b/agent_eval/_bootstrap.py
@@ -89,7 +89,7 @@ def _activate():
     if os.environ.get(_SENTINEL):
         return
 
-    pkg_dir = os.path.dirname(os.path.abspath(__file__))
+    pkg_dir = os.path.dirname(os.path.realpath(__file__))
     plugin_root = os.path.dirname(pkg_dir)
     venv_python, site_dirs = _venv_python_and_site(plugin_root)
 
--- a/skills/eval-analyze/scripts/validate_eval.py
+++ b/skills/eval-analyze/scripts/validate_eval.py
@@ -18,6 +18,25 @@ import yaml
 sys.path.insert(0, str(Path(__file__).parent))
 from find_skills import find_skill
 
+# Variables the scoring renderer always injects into a judge template. Must stay
+# equal to the kwargs of the render() call at the end of
+# skills/eval-run/scripts/score.py::_render_jinja2_template — accepting a name
+# score.py doesn't pass means a judge prompt validates here and then renders as a
+# blank section at scoring time (score.py uses a logging Undefined, not
+# StrictUndefined, so nothing errors). tests/test_validate_eval.py pins the two
+# together so the lists can't drift again.
+STANDARD_TEMPLATE_VARS = {
+    "inputs",           # case input.yaml, pre-rendered as text
+    "annotations",      # dataset annotations (dict + formatted __str__)
+    "annotations_text", # formatted annotation text for display
+    "conversation",     # root-level assistant text from events
+    "tool_trace",       # chronological trace of tool calls (Read, Bash, etc.)
+    "outputs",          # collected output files/events proxy
+    "arguments",        # judge arguments from eval.yaml
+    "reasoning",        # conversation including extended thinking
+    "evidence",         # lazily extracted verifiable evidence
+}
+
 
 def _extract_template_variables(template_text):
     """Extract undeclared root variable names from a Jinja2 template.
@@ -48,20 +67,7 @@ def _validate_template_variables(judges, outputs, dataset_schema, errors, warnin
     # Build set of available output names
     output_names = {o.get("name", "") for o in outputs if o.get("name")}
 
-    # Standard variables the scoring renderer (score._render_jinja2_template)
-    # always injects. Keep in sync with that render() call.
-    standard_vars = {
-        "inputs",           # Loaded from dataset input.yaml
-        "annotations",      # dataset annotations (dict + formatted __str__)
-        "annotations_text", # formatted annotation text for display
-        "conversation",     # root-level assistant text from events
-        "tool_trace",       # chronological trace of tool calls (Read, Bash, etc.)
-        "events",           # tool call log
-        "outputs",          # collected output files/events proxy
-        "arguments",        # judge arguments from eval.yaml
-        "reasoning",        # model reasoning content
-        "evidence",         # extracted judge evidence
-    }
+    standard_vars = STANDARD_TEMPLATE_VARS
 
     had_undefined_vars = False
 
@@ -119,8 +125,8 @@ def _validate_template_variables(judges, outputs, dataset_schema, errors, warnin
         errors.append(
             "\n💡 Template variable errors detected. Common fixes:\n"
             "   • Ensure all {{ variable }} references match output names or standard variables\n"
-            "   • Standard variables (always available): inputs, annotations, annotations_text, "
-            "conversation, tool_trace, events, outputs, arguments, reasoning, evidence\n"
+            f"   • Standard variables (always available): "
+            f"{', '.join(sorted(STANDARD_TEMPLATE_VARS))}\n"
             "   • For custom variables, add them to outputs section in eval.yaml\n"
             "   • Check dataset.schema documents expected structure of input.yaml and annotations.yaml"
         )
@@ -141,9 +147,11 @@ def _test_render_judge_templates(judges, outputs, errors, warnings):
     env.filters["tojson"] = lambda v: _json.dumps(v, indent=2, default=str)
 
     output_names = {o.get("name", "") for o in outputs if o.get("name")}
+    # One entry per STANDARD_TEMPLATE_VARS name (pinned by tests/test_validate_eval.py),
+    # each shaped like what score.py actually injects — `inputs` is pre-rendered
+    # text, not a dict, so `{{ inputs.field }}` is a mistake the render surfaces.
     mock_data = {
         "conversation": "Mock conversation",
-        "events": [],
         "inputs": "**prompt**: Mock prompt",
         "annotations": {"category": "test", "expected_files": []},
         "annotations_text": "- **category**: test",
@@ -195,8 +203,7 @@ def _test_render_judge_templates(judges, outputs, errors, warnings):
             "  1. Ensure all {{ variable }} references match output names\n"
             "  2. For dataset files (input.yaml, annotations.yaml), verify they're loaded by scoring\n"
             "  3. Check dataset.schema documents the expected structure\n"
-            "  4. Standard variables: inputs, annotations, annotations_text, "
-            "conversation, events, outputs, arguments, reasoning, evidence, tool_trace"
+            f"  4. Standard variables: {', '.join(sorted(STANDARD_TEMPLATE_VARS))}"
         )
 
 
--- a/skills/eval-analyze/scripts/validate_eval.py
+++ b/skills/eval-analyze/scripts/validate_eval.py
@@ -51,7 +51,7 @@ def _validate_template_variables(judges, outputs, dataset_schema, errors, warnin
     # Standard variables the scoring renderer (score._render_jinja2_template)
     # always injects. Keep in sync with that render() call.
     standard_vars = {
-        "input",            # Loaded from dataset input.yaml
+        "inputs",           # Loaded from dataset input.yaml
         "annotations",      # dataset annotations (dict + formatted __str__)
         "annotations_text", # formatted annotation text for display
         "conversation",     # root-level assistant text from events
@@ -59,6 +59,8 @@ def _validate_template_variables(judges, outputs, dataset_schema, errors, warnin
         "events",           # tool call log
         "outputs",          # collected output files/events proxy
         "arguments",        # judge arguments from eval.yaml
+        "reasoning",        # model reasoning content
+        "evidence",         # extracted judge evidence
     }
 
     had_undefined_vars = False
@@ -96,11 +98,11 @@ def _validate_template_variables(judges, outputs, dataset_schema, errors, warnin
             )
 
         # Check for common mistakes with input/annotations
-        if "input" in vars_used:
+        if "inputs" in vars_used:
             # Warn if input.yaml expected fields aren't documented
             if not dataset_schema or "input.yaml" not in dataset_schema:
                 warnings.append(
-                    f"judges.{name} uses {{{{ input }}}} but dataset.schema doesn't document input.yaml structure. "
+                    f"judges.{name} uses {{{{ inputs }}}} but dataset.schema doesn't document input.yaml structure. "
                     f"Document expected fields (prompt, expected_api, expected_documentation, etc.)"
                 )
 
@@ -117,8 +119,8 @@ def _validate_template_variables(judges, outputs, dataset_schema, errors, warnin
         errors.append(
             "\n💡 Template variable errors detected. Common fixes:\n"
             "   • Ensure all {{ variable }} references match output names or standard variables\n"
-            "   • Standard variables (always available): input, annotations, annotations_text, "
-            "conversation, tool_trace, events, outputs, arguments\n"
+            "   • Standard variables (always available): inputs, annotations, annotations_text, "
+            "conversation, tool_trace, events, outputs, arguments, reasoning, evidence\n"
             "   • For custom variables, add them to outputs section in eval.yaml\n"
             "   • Check dataset.schema documents expected structure of input.yaml and annotations.yaml"
         )
@@ -142,11 +144,14 @@ def _test_render_judge_templates(judges, outputs, errors, warnings):
     mock_data = {
         "conversation": "Mock conversation",
         "events": [],
-        "input": {"prompt": "Mock prompt", "expected_documentation": []},
+        "inputs": "**prompt**: Mock prompt",
         "annotations": {"category": "test", "expected_files": []},
         "annotations_text": "- **category**: test",
         "arguments": {},
         "outputs": {"files": {}, "events": [], "conversation": ""},
+        "reasoning": "Mock reasoning",
+        "evidence": "Mock evidence",
+        "tool_trace": "",
     }
     # Add all declared outputs
     for name in output_names:
@@ -190,8 +195,8 @@ def _test_render_judge_templates(judges, outputs, errors, warnings):
             "  1. Ensure all {{ variable }} references match output names\n"
             "  2. For dataset files (input.yaml, annotations.yaml), verify they're loaded by scoring\n"
             "  3. Check dataset.schema documents the expected structure\n"
-            "  4. Standard variables: input, annotations, annotations_text, "
-            "conversation, events, outputs, arguments"
+            "  4. Standard variables: inputs, annotations, annotations_text, "
+            "conversation, events, outputs, arguments, reasoning, evidence, tool_trace"
         )
 
 
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -233,19 +233,25 @@ def test_bootstrap_symlink_realpath(tmp_path):
     symlink_path = skills_dir / "agent_eval"
     os.symlink(root / "agent_eval", symlink_path, target_is_directory=True)
     
-    code = textwrap.dedent("""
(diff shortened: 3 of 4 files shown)


tests/test_bootstrap.py
```diff
@@ -223,3 +223,39 @@ def test_true_script_under_abi_mismatch_execs(tmp_path):
...
+    
+    env = dict(os.environ)
+    env.pop("_AGENT_EVAL_BOOTSTRAP_DONE", None)
+    env.pop("PYTHONPATH", None)  # Ensure it doesn't bypass the symlink
+    
+    proc = subprocess.run([sys.executable, str(script)], cwd=str(skills_dir), env=env,
+                          capture_output=True, text=True, timeout=120)
+    
+    assert proc.returncode == 0, f"child failed:\n{proc.stdout}\n{proc.stderr}"
+    assert "DONE=1" in proc.stdout, "Venv was not activated (SENTINEL not set)"
```
astefanutti: **The new test passes with the fix reverted — it gives the `realpath` change zero regression coverage.**

`_activate()` sets `_AGENT_EVAL_BOOTSTRAP_DONE` on *every* path, including the "no venv found" early return at `agent_eval/_bootstrap.py:97-99` — which is exactly the branch the buggy `abspath` code takes. So `DONE=1` cannot distinguish fixed from broken, and the failure message ("Venv was not activated") is misleading.

The companion assertion at line 244 (`'skills' in b.__file__`) is equally branch-insensitive: CPython never realpaths `__file__` on import, so it just re-asserts the test's own symlink setup.

Verified empirically — reverting line 92 to `os.path.abspath(__file__)` on this branch leaves the whole bootstrap suite green:

```
$ sed -i 's/os.path.realpath(__file__)/os.path.abspath(__file__)/' agent_eval/_bootstrap.py
$ python3 -m pytest tests/test_bootstrap.py -q
14 passed in 0.16s
```

Assert an observable that actually differs between the two branches. In the child script:

```diff
-        # _bootstrap already ran on import, check if it successfully found .eval-venv
-        # by verifying the sentinel was set
-        print("DONE=" + str(os.environ.get('_AGENT_EVAL_BOOTSTRAP_DONE')))
+        # Only the realpath resolution finds <root>/.eval-venv from the symlinked
+        # package dir, so its site-packages landing on sys.path is the discriminator.
+        print("VENV_ON_PATH=" + str(any('.eval-venv' in p for p in sys.path)))
```

and at line 261:

```diff
-    assert "DONE=1" in proc.stdout, "Venv was not activated (SENTINEL not set)"
+    assert "VENV_ON_PATH=True" in proc.stdout, \
+        f"venv site-packages not on sys.path:\n{proc.stdout}"
```

I ran both directions to confirm it discriminates:

```
--- with abspath (bug present) ---   1 failed
--- with realpath (fix present) ---  1 passed
```

This also restores the convention the file sets for itself — `test_true_script_under_abi_mismatch_execs` (lines 219-225) exists specifically as a "positive control proving the -m test above is not passing vacuously", and `526ba2c` replaced an earlier bootstrap test for exactly this reason.

Worth tightening line 244 to the full `os.path.join("skills", "eval-analyze", "scripts")` sub-path too, so a tmp dir that happens to contain "skills" can't pass spuriously.

astefanutti (commented):
## Review

Both production fixes are correct. Requesting one change: the test added to guard the `realpath` fix is vacuous — it passes unchanged with the fix reverted.

### The venv change is consistent with every prior venv PR

I walked each venv-touching commit and checked the invariant it established against the `abspath` → `realpath` swap:

| Commit | Invariant | Effect |
|---|---|---|
| `dd5c76e` | venv at `<plugin_root>/.eval-venv` holds only third-party deps; `agent_eval` resolves from the source tree | untouched — only changes *which* dir is `plugin_root` |
| `77d89ca` / `be6afb6` | every script imports `_bootstrap` first | untouched |
| `9d0fca6` / `526ba2c` | never `execv` on `-m`/`-c`/`exec_module` (the harbor duplicate-pod bug) | untouched; the sentinel is still set before `execv` and inherited, so the re-exec stays idempotent |
| `df134af` / `617f62f` | truststore injection runs *after* activation | untouched — ordering at `_bootstrap.py:158-159` unchanged |
| `c978627` | flexible eval-dir layout | orthogonal |

No `.eval-venv` create/lookup split is introduced: `scripts/ensure_deps.py:28` uses an unresolved `Path(__file__).parent.parent` and the SKILL.md snippets use `${CLAUDE_SKILL_DIR}/../../.eval-venv`, but `scripts/ensure_deps.py` is a real file, and where the plugin root is itself a symlink the kernel resolves it to the same physical directory. Both still land on the same venv. I could not construct a supported layout — symlinked plugin root, macOS `/tmp`→`/private/tmp`, container site-packages install, `pip install -e` — where `realpath` is worse than `abspath`.

### Summary

| | |
|---|---|
| 🔴 Change requested | `tests/test_bootstrap.py:261` — vacuous regression test |
| 🟡 Minor | `validate_eval.py:59` — `events` survived the sync |
| 🟡 Minor | `validate_eval.py:147` — the `validate_eval` half ships with no tests |
| 🟢 | `_bootstrap.py:92` — correct, and aligns the last holdout with 5 other call sites |

Not issues, for the record: `{{ input.* }}` in README / CLAUDE.md / website is **correct** and must not be migrated — `config.resolve_arguments()` still binds `input` for `execution.arguments`/`execution.prompt`; `inputs` is judge-only, and the PR correctly scoped its change to `_validate_template_variables`, which only sees judges. CI is green and there is no ruff/flake8 in the repo, so the trailing whitespace in the new test won't fail lint.

agent_eval/_bootstrap.py
```diff
@@ -89,7 +89,7 @@ def _activate():
     if os.environ.get(_SENTINEL):
         return
 
-    pkg_dir = os.path.dirname(os.path.abspath(__file__))
+    pkg_dir = os.path.dirname(os.path.realpath(__file__))
```
astefanutti, line 92: ✅ Correct, and this is a convergence rather than a one-off.

`_bootstrap` was the **last holdout** still using `abspath`. Every other repo-root-from-`__file__` computation already resolves symlinks:

- `agent_eval/tools/interception.py:35`
- `agent_eval/harbor/run.py:31`
- `agent_eval/harbor/reward.py:176`
- `agent_eval/evalhub/runner.py:36`
- `skills/eval-anova/scripts/orchestrate.py:63` — whose comment literally cites *"the `agent_eval` symlink some test/plugin layouts use"*

So this one-liner both fixes the bug and removes the inconsistency. Confirmed end to end for the 7 skills that ship a `scripts/agent_eval` symlink:

```
$ PYTHONPATH=skills/eval-setup/scripts python3 -c "import agent_eval._bootstrap, sys; \
    print([p for p in sys.path if '.eval-venv' in p])"
```

→ `[]` before, `['<root>/.eval-venv/lib/python3.x/site-packages']` after.

skills/eval-analyze/scripts/validate_eval.py
```diff
@@ -51,14 +51,16 @@ def _validate_template_variables(judges, outputs, dataset_schema, errors, warnin
     # Standard variables the scoring renderer (score._render_jinja2_template)
     # always injects. Keep in sync with that render() call.
     standard_vars = {
-        "input",            # Loaded from dataset input.yaml
+        "inputs",           # Loaded from dataset input.yaml
         "annotations",      # dataset annotations (dict + formatted __str__)
         "annotations_text", # formatted annotation text for display
         "conversation",     # root-level assistant text from events
         "tool_trace",       # chronological trace of tool calls (Read, Bash, etc.)
         "events",           # tool call log
```
astefanutti: `events` survived the sync — it's the one entry that still doesn't match the renderer.

`score._render_jinja2_template` (`skills/eval-run/scripts/score.py:706-716`) passes exactly:

```
arguments, outputs, annotations, annotations_text, conversation,
reasoning, inputs, evidence, tool_trace
```

so after this PR `standard_vars - score_vars == {"events"}`. It's only reachable as `outputs.events`, never at top level.

Low impact — `score.py` uses `make_logging_undefined(base=Undefined)`, not `StrictUndefined`, so a judge writing `{{ events }}` validates fine and then renders as a silently blank section at scoring time rather than erroring. But it's precisely the drift this PR set out to eliminate, one line from the lines it touched.

Drop it from `standard_vars`, from the two guidance strings (lines ~122 and ~198) and from `mock_data` (line 146) — or land the CLAUDE.md *Remaining Work* item that actually injects `outputs["events"]` as a root-level var.

skills/eval-analyze/scripts/validate_eval.py
```diff
@@ -142,11 +144,14 @@ def _test_render_judge_templates(judges, outputs, errors, warnings):
     mock_data = {
         "conversation": "Mock conversation",
         "events": [],
-        "input": {"prompt": "Mock prompt", "expected_documentation": []},
+        "inputs": "**prompt**: Mock prompt",
```
astefanutti, line 155: Making `inputs` a plain string is right — `score.py:689` does `out.get("inputs", "")`, so this now matches the renderer.

Two follow-ons worth considering, neither blocking:

1. Because the mock is a string, a judge writing `{{ inputs.prompt }}` no longer errors in `_test_render_judge_templates` — it renders empty, and does the same at real scoring time (non-strict `Undefined`). The dict mock used to surface that class of typo. Pre-existing shape rather than something this PR introduces, but the change makes it a bit more likely.

2. **This half of the PR has no test coverage at all.** The behaviour change is real — on `main` a judge prompt containing `{{ inputs }}` is hard-rejected with *"references undefined template variable(s)"*; on this branch it validates cleanly — yet `tests/` exercises neither `_validate_template_variables` nor `_test_render_judge_templates` nor `validate_config` (the only reference anywhere is an `.exists()` assertion at `tests/test_prompt_analysis.py:286`).

   The list at line 53 carries a *"Keep in sync with that render() call"* comment and has already drifted once. A ~10-line guard would stop the next drift: extract the set to a module-level `STANDARD_TEMPLATE_VARS`, assert it equals the kwargs `score.py:706-716` actually passes, and add two `validate_config()` cases on tmp fixtures (`{{ inputs }}` → no error, `{{ input }}` → error).

astefanutti (changes requested):
**Requesting changes** on one item — formalising the ask from [my review above](https://github.com/opendatahub-io/agent-eval-harness/pull/179#pullrequestreview-4904878561) (both production fixes are correct; this is purely about the test).

🔴 **[`tests/test_bootstrap.py:261`](https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3756847321) — the new test passes with the fix reverted.**

`_activate()` sets `_AGENT_EVAL_BOOTSTRAP_DONE` on every path, including the "no venv found" early return at `agent_eval/_bootstrap.py:97-99` — the exact branch the buggy `abspath` code takes. So `DONE=1` can't tell fixed from broken:

```
$ sed -i 's/os.path.realpath(__file__)/os.path.abspath(__file__)/' agent_eval/_bootstrap.py
$ python3 -m pytest tests/test_bootstrap.py -q
14 passed in 0.16s
```

The one-line swap to `print("VENV_ON_PATH=" + str(any('.eval-venv' in p for p in sys.path)))` (full diff in the inline thread) fails under `abspath` and passes under `realpath` — I ran it both ways.

The other three threads are non-blocking: [`_bootstrap.py:92`](https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3756847326) is a 🟢 confirmation that the venv change is consistent with every prior venv PR, and the two `validate_eval.py` threads ([`:59`](https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3756847336), [`:147`](https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3756847343)) are minors. Happy to re-review as soon as the test discriminates.

astefanutti (commented):
Thanks @RedTanny, I've left a couple of comments otherwise looks good to me.

RedTanny:
Hi @astefanutti 
thanks for reviewing 
added a fixed  (https://github.com/opendatahub-io/agent-eval-harness/pull/179/changes/44cb76f0fe5f6a0540d52fdebd7dc27a175d0ccf)  for requested change on the test file 🔴 https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3756847321

astefanutti (approved):
Thanks @RedTanny — `44cb76f0` fixes the blocker. **Approving.**

I re-verified the regression guard rather than taking it on trust, and it now discriminates: `14 passed` with the fix, and with `agent_eval/_bootstrap.py:92` reverted to `abspath` the child reports `VENV_ON_PATH=False` and the test fails. That's real protection where there was none.

### One thing left, non-blocking

[CodeRabbit's point](https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3757362806) about the substring match is **correct** — I confirmed it empirically. `any('.eval-venv' in p for p in sys.path)` also matches the *test runner's own* venv, so if you run the suite with `.eval-venv/bin/python3` the child inherits that interpreter's site-packages and the check is true no matter what:

| runner | `_bootstrap.py:92` | substring (current) | exact path (proposed) |
|---|---|---|---|
| system python | `abspath` (bug) | ❌ fails | ❌ fails |
| system python | `realpath` (fix) | ✅ passes | ✅ passes |
| `.eval-venv/bin/python3` | `abspath` (bug) | ⚠️ **passes** | ❌ fails |
| `.eval-venv/bin/python3` | `realpath` (fix) | ✅ passes | ✅ passes |

My earlier verification used the system interpreter, which is why I didn't catch this the first time — the guard is sound there, and it's sound in CI (`tests.yml` uses `setup-python 3.12` + `pip install -e ".[all,test]"`, never `.eval-venv`). So **CI does catch a revert** and this doesn't block the merge. It only bites someone running the suite locally with the harness's own venv, which is a plausible enough habit to be worth closing.

Asserting the exact fake `site-packages` path fixes all four cells. Tested both interpreters × both bootstrap variants:

```diff
-    code = textwrap.dedent("""
+    site = str(root / ".eval-venv" / "lib" / f"python{_RUNNING}" / "site-packages")
+    code = textwrap.dedent(f"""
         import sys
         import os
         # Insert scripts dir at front so we import the symlink, not the real package
         sys.path.insert(0, os.path.dirname(__file__))
         import agent_eval._bootstrap as b

         # Verify it was loaded via the symlink
         assert 'skills' in b.__file__ or 'eval-analyze' in b.__file__

-        # _bootstrap already ran on import, check if it successfully found .eval-venv
-        # by verifying that the venv's site-packages was appended to sys.path
-        print("VENV_ON_PATH=" + str(any('.eval-venv' in p for p in sys.path)))
+        # Only realpath resolution reaches <root>/.eval-venv from the symlinked
+        # package dir. Compare the EXACT path: a substring test also matches the
+        # *runner's* own venv when the suite is run with .eval-venv/bin/python3.
+        want = {site!r}
+        print("VENV_ON_PATH=" + str(any(
+            os.path.realpath(p) == os.path.realpath(want) for p in sys.path)))
     """)
```

`os.path.realpath` on both sides handles macOS `/tmp` → `/private/tmp`. Full suite stays green (`854 passed, 2 skipped`).

Two cosmetic nits while you're in there, both optional:
- The failure message has a literal `\\n` rather than a newline: `f"...not in sys.path)\\nstdout: {proc.stdout}"`.
- `assert 'skills' in b.__file__` could be the full `os.path.join("skills", "eval-analyze", "scripts")` sub-path, so a tmp dir that happens to contain "skills" can't satisfy it.

### Still open from the first review (both minor, your call)

- [`validate_eval.py:59`](https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3756847336) — `events` is still in `standard_vars` but `score.py` never injects it. One line from the lines this PR already touched.
- [`validate_eval.py:147`](https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3756847343) — the `validate_eval` half still has no test coverage.

Neither blocks. Happy for this to go in as-is and the exact-path tweak to follow, or to fold it in here — whichever you prefer.

astefanutti:
Pushed `15dd16e` to this branch directly (maintainer-modify) rather than leaving you a list — all of it is review follow-up, no change to the two fixes you wrote. Revert or amend freely.

**`tests/test_bootstrap.py`** — the substring check had a residual false-pass that [CodeRabbit spotted](https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3757362806) and I confirmed: `any('.eval-venv' in p for p in sys.path)` also matches the *runner's own* venv, so running the suite with `.eval-venv/bin/python3` passed even with the bug present. Now asserts the exact fake `site-packages` path. Verified all four combinations:

| runner | `_bootstrap.py:92` | before | after |
|---|---|---|---|
| system python | `abspath` | ❌ fails | ❌ fails |
| system python | `realpath` | ✅ passes | ✅ passes |
| `.eval-venv/bin/python3` | `abspath` | ⚠️ passed | ❌ fails |
| `.eval-venv/bin/python3` | `realpath` | ✅ passes | ✅ passes |

Also folded in the two cosmetic nits — the literal `\n` in the failure message, and `'skills' in b.__file__` → the full `skills/eval-analyze/scripts` sub-path.

**`validate_eval.py`** — dropped `events`, the last entry still out of sync. `score._render_jinja2_template` injects `arguments, outputs, annotations, annotations_text, conversation, reasoning, inputs, evidence, tool_trace` and never a root-level `events` (only `outputs.events`). Accepting it meant a judge prompt validated here and then rendered a blank section at scoring time, because score.py uses a logging `Undefined` rather than `StrictUndefined` — silent, which is the worst version. Nothing in the repo used it.

Hoisted the set to a module-level `STANDARD_TEMPLATE_VARS` and generated both "Standard variables:" hints from it, so the guidance can no longer advertise a name the validator rejects.

**`tests/test_validate_eval.py`** (new) — this file had no coverage at all, which is why it drifted twice. The key test AST-extracts the kwargs of score.py's `render()` call and asserts set equality with `STANDARD_TEMPLATE_VARS`, so neither can move without the other; I checked it fails when a phantom var is added. Plus end-to-end cases: `{{ inputs }}`/`{{ reasoning }}`/`{{ evidence }}`/`{{ tool_trace }}` clean, `{{ input }}` and `{{ events }}` rejected.

Full suite `865 passed, 2 skipped`. My approval stands — merge when you're happy with it.

One note for whoever merges: **#179 should go in before #180**, since #180's venv-activation half is inert until this `realpath` fix lands.

RedTanny:
Thanks @astefanutti for the follow-up — looks good to me. I don't have merge permissions on this repo, so please merge when you can.

astefanutti (approved):
Thanks @RedTanny!

PR #180, opened 2026-08-11
fix(venv): activate the venv in eval-anova/check/compare scripts

astefanutti:
## Description

Follow-up to the adjacent gaps found while reviewing #179.

Three skills — **eval-anova**, **eval-check**, **eval-compare** — shipped no `scripts/agent_eval` symlink and no `agent_eval._bootstrap` import, so they never activated `.eval-venv`. For eval-anova this wasn't just a missing venv. Invoked the documented way, `sys.path[0]` is the script's own directory, so the import failed outright:

```console
$ cd /tmp && python3 <repo>/skills/eval-anova/scripts/orchestrate.py --help
ModuleNotFoundError: No module named 'agent_eval'
```

Same class of bug in four more entry points found while fixing it.

### What changed

| Area | Change |
|---|---|
| Symlinks | `scripts/agent_eval -> ../../../agent_eval` for eval-anova, eval-check, eval-compare (matching the 7 skills that already had it) |
| Bootstrap import | `orchestrate.py`, `analyze.py`, `design.py`, `harness_inventory.py`, `reference_checker.py`, `compare.py` |
| Ordering | `generate_synthetic.py` imported `yaml` *before* `_bootstrap`, so pyyaml resolved against the system interpreter |
| Also missing | `scripts/discover.py` (invoked by six SKILL.md files) and `skills/eval-analyze/scripts/reorganize.py` — both reach `agent_eval` via a `sys.path` insert but never activate the venv |
| In-package entry points | `agent_eval/state.py` (invoked directly by three skills, imports yaml) and `agent_eval/cli/trace_run.py` (the `claude-trace` console script) |

`eval-check` imports `yaml` lazily *inside functions*, so it needs the module-level bootstrap even though its top-level imports are all stdlib.

`state.py` can't use a plain `import agent_eval._bootstrap` — SKILL.md invokes it as `python3 ${CLAUDE_SKILL_DIR}/scripts/agent_eval/state.py`, which puts `sys.path[0]` *inside* the package, so `import agent_eval` doesn't resolve. Hence the `__package__` guard that puts the plugin root on the path first.

`eval-anova/scripts/report.py`, `resolve_prompt.py`, `otel-push.py` and `generate-test-trace.py` are stdlib-only and are left alone.

### Sentinel leak into child processes

Making orchestrate.py bootstrap introduced a regression that this PR also fixes. eval-anova runs each eval-run step as a `python3 <step>.py` child inheriting `os.environ`. The bootstrap sentinel is designed to survive `os.execv` *within* one process — if it crosses into a child, the child short-circuits `_activate()` and loses the venv's site-packages.

That bites in the common ABI-**match** case: the parent patches only its own `sys.path`, and `sys.executable` is still the unpatched interpreter. `_child_env()` strips the sentinel for both spawn sites.

### Relationship to #179

Complementary, no overlap — `agent_eval/_bootstrap.py` is untouched here.

- **This PR alone** fixes the `ModuleNotFoundError`. `discover.py`, `compare.py`, `reference_checker.py` and `design.py` all run clean afterwards.
- **The venv-activation half additionally needs #179.** Bootstrap resolves `plugin_root` from `__file__` through the new symlink, so with `abspath` it lands on `skills/<skill>/scripts` and finds no venv. With #179's `realpath` one-liner applied locally, `orchestrate.py`, `analyze.py` and `harness_inventory.py` also run clean under system python 3.9 — bootstrap re-execs into the venv's 3.11.

Merging in either order is safe.

## How Has This Been Tested?

`tests/test_venv_activation.py` guards all three invariants structurally: every `skills/*/scripts/` dir carries the symlink; `_bootstrap` precedes the first third-party import in any entry point that has one (counting deferred imports inside functions, and covering `agent_eval` modules with a `__main__` block); and the sentinel never crosses into a child env.

- **Fails 15 times on the parent commit**, passes here — verified in a clean worktree, not just in aggregate.
- Full suite: `898 passed, 8 skipped`.
- Runtime: each affected script re-run under system python 3.9 with a clean env, before and after, and again with #179 applied.
- `scripts/ensure_deps.py` is exempt and documented as such — it *creates* the venv, so it cannot activate it.
- Confirmed the new eval-check symlink doesn't perturb eval-check's own scanners: `harness_inventory` output is byte-identical across the two trees, and `reference_checker` reports zero broken references / missing scripts / orphan skills on both.

## Known follow-up (not in this PR)

`.eval-venv` never receives the `[anova]` extras (scipy, statsmodels, pandas, pingouin) — `ensure_deps.py` installs only pyyaml/mlflow/anthropic/jinja2, while SKILL.md tells users `pip install -e ".[anova]"`, which lands elsewhere. So under an ABI *mismatch*, `analyze.py` now re-execs into a venv without pandas. Teaching `ensure_deps._resolve_deps` to add the anova extras when a `matrix:` block is present is the natural fix, but it pulls a ~200MB install into SessionStart and deserves its own PR.

## Merge criteria:

- [x] Testing instructions have been added in the PR body.

## Summary by CodeRabbit

- **New Features**
  - Evaluation and analysis commands now automatically activate the project’s configured Python environment.
  - Supported tools run consistently from their available script locations.
  - Child processes receive a clean runtime environment.
  - Missing optional analysis dependencies now provide clearer, interpreter-specific installation guidance.

- **Bug Fixes**
  - Improved reliability for direct tool execution, configuration discovery, state inspection, comparison, and orchestration workflows.

- **Documentation**
  - Added guidance for creating scripts and skills with the required environment setup.

- **Tests**
  - Added coverage for environment activation, script setup, dependency guidance, and child-process handling.


astefanutti:
## Addressing the CodeRabbit review

**1 & 2 — `_bootstrap.py:92` should use `realpath`/`resolve()` instead of `abspath`** (comments on `skills/eval-anova/scripts/agent_eval` and `analyze.py:17`)

Correct, and both comments are describing the same root cause. That fix is **#179**, already open and reviewed — this PR deliberately does not touch `agent_eval/_bootstrap.py`, so the two don't conflict and can merge in either order.

The "Relationship to #179" section of the PR description covers this: the `ModuleNotFoundError` is fixed by *this* PR alone, and the venv-activation half lands once #179 does. Verified by applying #179's one-liner locally — `orchestrate.py`, `analyze.py` and `harness_inventory.py` then all run clean under system python 3.9, with bootstrap re-execing into the venv's 3.11.

On the request to "add an execution-level regression test for a skill symlink" — that belongs with the fix, and I have [asked for exactly that on #179](https://github.com/opendatahub-io/agent-eval-harness/pull/179#discussion_r3756847321): the test it currently ships passes with the fix reverted, and I gave a tested one-line replacement that discriminates.

**3 — `_child_env()` restores the sentinel if `extra` contains it** (`orchestrate.py:73`)

Valid, fixed. `extra` is now applied first and the sentinel popped afterwards, so an override cannot reinstate it:

```python
env = dict(os.environ)
if extra:
    env.update(extra)
# After the overrides, so `extra` cannot put the sentinel back.
env.pop(agent_eval._bootstrap._SENTINEL, None)
```

Added the suggested `_child_env({_SENTINEL: "1"})` assertion, and confirmed it discriminates — `True` under the old ordering, `False` under the new one. No live caller passes the sentinel today (`extra_env` only ever carries `AGENT_EVAL_RUNS_DIR`), so this was latent, but the ordering is the right way round.

---

Also added, prompted by this whole class of bug: a short **"Adding a skill"** section in `CLAUDE.md` documenting the two requirements (the `scripts/agent_eval` symlink and the first-import bootstrap), the exemptions, and the child-process sentinel rule — so the next skill doesn't rediscover them through a CI failure. `tests/test_venv_activation.py` remains the mechanical enforcement.

Full suite: `898 passed, 8 skipped`.

astefanutti:
@coderabbitai review

astefanutti:
Rebased onto main now that #179 is in (`c6358f5`), and re-verified the claim in the description **for real** rather than by simulating the `realpath` one-liner.

**Every affected script, system python 3.9, clean env, documented invocation** — all six now run:

| script | before | after |
|---|---|---|
| `eval-anova/orchestrate.py` | `ModuleNotFoundError: agent_eval` | ✅ |
| `eval-anova/analyze.py` | `ModuleNotFoundError: pandas` | ✅ |
| `eval-anova/design.py` | `ModuleNotFoundError: agent_eval` | ✅ |
| `eval-compare/compare.py` | ran, no venv | ✅ |
| `eval-check/harness_inventory.py` | `TypeError: unsupported operand \|` (PEP 604 under 3.9) | ✅ |
| `eval-check/reference_checker.py` | ran, no venv | ✅ |

Plus `agent_eval/state.py` through the symlink path and `scripts/discover.py`.

**Bootstrap really does re-exec on an ABI mismatch** — launched with system 3.9, the script continues under `.eval-venv/bin/python3` with pandas 2.3.3 available. That's also what fixes the `dict | None` syntax error in eval-check without touching the annotations.

**The child-process sentinel fix is load-bearing**, and now on the real path. Proved it with a controlled interpreter rather than a convenient one — a bare 3.11 venv with no yaml, matching `.eval-venv`'s ABI so the parent takes the `_patch_syspath` branch (the common case, since `ensure_deps.py` builds the venv from `sys.executable`):

```
control OK: bare interpreter has no yaml
parent activated the venv, yaml = 6.0.3
  WITHOUT fix (sentinel inherited) -> rc=1 | ModuleNotFoundError: No module named 'yaml'
  WITH    _child_env()             -> rc=0 | child got yaml 6.0.3
```

My first attempt at this used an interpreter that happened to have yaml installed, so it passed either way and proved nothing — worth noting since that is exactly the failure mode we spent #179 fixing.

Full suite on the rebase: `914 passed, 8 skipped`.

astefanutti:
Both fixed in `6ba5852`. The first one is a good catch with some irony — that message exists specifically to name the right interpreter, and it named the wrong one.

**`stats/__init__.py:47` — install target.** Confirmed: `sys.executable` is only correct on an ABI *mismatch*, where `_bootstrap` re-execs and `sys.executable` already *is* the venv python. On the common ABI-*match* path it just patches `sys.path`, so `sys.executable` stays the launcher. Reproduced under a 3.11 launcher with a 3.11 venv:

```
  interpreter: /Library/Frameworks/Python.framework/Versions/3.11/bin/python3
  --> .eval-venv in the command? False
```

Worth spelling out why that's harmful rather than merely untidy: installing into the launcher *appears* to work, because the launcher's site-packages is still on `sys.path` behind the venv's. It then silently vanishes the moment someone runs under a different python and bootstrap re-execs into the venv.

`_installer_python()` now resolves the venv exactly as bootstrap does (`_venv_python_and_site`), falling back to the running interpreter only when there's no venv — harbor verifier, evalhub pod. When the two differ the message reports both:

```
  missing:      one or more of them
  running:      /Library/Frameworks/Python.framework/Versions/3.11/bin/python3
  imports from: /Users/.../agent-eval-harness/.eval-venv/bin/python3   <- the extra has to go here
```

**`analyze.py:129` — lost module name.** Also correct. `agent_eval.anova.stats` swallowed the `ImportError`, so the `ANOVA_AVAILABLE` gate had nothing to pass and reported `one or more of them` even when it knew scipy was missing. Recorded as `_IMPORT_ERROR` and used when the caller supplies none.

Both are pinned by tests that fail against the previous behaviour — verified by reverting each:

```
  with _installer_python reverted to sys.executable: 1 failed, 3 passed
  with the recorded ImportError dropped:             1 failed
```

One deliberate detail: `test_installer_target_is_the_venv_when_one_exists` builds a synthetic `.eval-venv` under `tmp_path` rather than relying on the repo's. CI does `pip install -e ".[all,test]"` and never creates `.eval-venv`, so a test comparing against the real one would pass trivially there — the synthetic root keeps it discriminating in CI.

Full suite: `917 passed, 8 skipped`.

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
Fixed in `0d68050`. The finding is right, and there was a plainer bug sitting underneath it that's worth naming.

**The interpreter path wasn't quoted at all.** Not an injection issue — the command was simply broken for anyone whose checkout lives under a directory with a space:

```
generated:   /Users/me/My Code/repo/.eval-venv/bin/python3 -m pip install -e "/Users/me/My Code/repo[anova]"
shlex.split: ['/Users/me/My', 'Code/repo/.eval-venv/bin/python3', '-m', 'pip', ...]
```

argv[0] truncates at the space. On macOS that's a common enough layout to hit in practice.

**The injection point is real but narrow.** `$(...)` and backticks do survive double quotes, so a checkout at `/tmp/$(id)/repo` would execute on paste. I'd rate it lower than Major though: it needs an attacker who already controls the path your checkout sits at — and if they have that, they can edit the Python you're about to run, which doesn't require you to paste anything. Worth fixing as hygiene, not because it's a plausible attack path.

`shlex.quote` on both arguments handles all of it, and incidentally protects the `[anova]` suffix from zsh glob expansion — which is why the double quotes were there to begin with:

```
/Users/.../.eval-venv/bin/python3 -m pip install -e '/Users/.../agent-eval-harness[anova]'
```

**On the test**: rather than asserting the `shlex.quote()` output as a string, I round-trip the printed line back through `shlex.split()` and assert the exact argv. That tests the property we actually care about — what the shell builds — instead of the escaping mechanism, so it stays valid if the quoting style changes. Parametrised over spaces, `$(id > /tmp/pwned)`, backticks, glob brackets, and a bare apostrophe. Verified the hostile cases fail against the previous f-string:

```
reverted to the unquoted f-string: 3 failed, 3 passed
```

Full suite: `922 passed, 8 skipped`.

PR #181, opened 2026-08-11
fix(deps): resolve venv deps from every eval.yaml, not just the first

astefanutti:
## Description

`scripts/ensure_deps.py` (the SessionStart hook that provisions `.eval-venv`) resolved its dependency set from **one** eval.yaml:

```python
configs = discover_configs(cwd)
if configs:
    return configs[0].path      # <- only the first
```

But `eval/<name>/eval.yaml` — several evals in one project — is a supported layout (spec 005), and sibling evals rarely need the same packages: one may log to MLflow while another uses LLM judges. Whichever config happened to sort first silently decided what got installed; the others then failed on an import of something that was never there.

**Reproduced** — a project with `eval/aaa-judge/` (LLM judge) and `eval/zzz-mlflow/` (mlflow block) resolves to:

```
['pyyaml>=6.0', 'anthropic[vertex]>=0.40', 'jinja2>=3.0']
```

mlflow dropped, purely because of the directory name.

### What changed

- **`_find_eval_yamls`** (renamed, plural) returns every discovered config.
- **`_resolve_deps`** unions the per-config deps, de-duplicating by spec. `_compute_stamp` already sorts, so discovery order doesn't churn the install cache.
- **`_deps_for_config`** extracts the per-config logic and never raises. Previously a single unparseable eval.yaml hit an early `return deps` that also dropped every config after it; now it contributes nothing and the scan continues.

The fallback stays first-match on purpose: it only runs when discovery itself failed, and `cwd/eval.yaml` vs `plugin_root/eval.yaml` are alternatives rather than siblings.

## How Has This Been Tested?

`tests/test_ensure_deps.py` is new — `ensure_deps.py` had **no test coverage at all**, which is largely why this went unnoticed. 14 tests covering per-config resolution, the union behaviour, de-duplication, parse-failure isolation, stamp order-independence, and the discovery fallback.

The union test fails on the parent commit for the right reason:

```
AssertionError: mlflow config was ignored — deps resolved from one config only
assert 'mlflow[genai]>=3.5' in ['pyyaml>=6.0', 'anthropic[vertex]>=0.40', 'jinja2>=3.0']
```

Full suite: `867 passed, 2 skipped`.

## Context

Came out of the discussion on #180 about whether to auto-install the `[anova]` extra gated on a `matrix:` block. That gate is defensible only once dep resolution stops depending on which config sorts first — a 272MB install that silently no-ops for half the projects that need it would be worse than no gate. This PR is that prerequisite; it does **not** add the anova gate, which remains an open design question (an explicit `dependencies:` opt-in may be the better shape).

Independent of #180 — no overlapping files.

## Merge criteria:

- [x] Testing instructions have been added in the PR body.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dependency setup now detects requirements across all available evaluation configurations.
  * Duplicate dependencies are removed automatically.
  * Invalid, unreadable, or missing configurations no longer interrupt dependency resolution.
  * Discovery retains fallback behavior when configuration scanning is unavailable and warns when dependencies may be missing.

* **Tests**
  * Added coverage for multi-configuration resolution, error handling, deduplication, caching, and fallback scenarios.

### DIFF
--- a/scripts/ensure_deps.py
+++ b/scripts/ensure_deps.py
@@ -48,12 +48,29 @@ def main():
 
 
 def _resolve_deps(plugin_root):
-    """Determine which deps are needed based on eval.yaml."""
-    deps = [("pyyaml>=6.0", "yaml")]
+    """Determine which deps are needed, as the union over every eval.yaml.
 
-    eval_yaml = _find_eval_yaml(plugin_root)
-    if not eval_yaml or not eval_yaml.exists():
-        return deps
+    A project can hold several evals (``eval/<name>/eval.yaml`` is a supported
+    layout), and they don't need the same packages — one may log to MLflow while
+    another uses LLM judges. Resolving against a single config left the others
+    importing packages that were never installed, with the outcome depending on
+    discovery order.
+    """
+    deps = [("pyyaml>=6.0", "yaml")]
+    seen = {spec for spec, _ in deps}
+
+    for eval_yaml in _find_eval_yamls(plugin_root):
+        for spec, module in _deps_for_config(eval_yaml):
+            if spec not in seen:
+                seen.add(spec)
+                deps.append((spec, module))
+    return deps
+
+
+def _deps_for_config(eval_yaml):
+    """Optional deps implied by one eval.yaml. Never raises — an unparseable or
+    unreadable config contributes nothing rather than sinking the whole scan."""
+    deps = []
 
     try:
         import yaml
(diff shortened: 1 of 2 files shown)


astefanutti:
@coderabbitai review

astefanutti:
Good catch, fixed in 3d8e630 — and it lands squarely on this PR's thesis: the whole point is to stop dep resolution silently under-installing, so leaving a silent `except Exception: pass` that degrades to a single config was the same bug one layer down.

Kept the broad handler — a discovery bug must never block SessionStart — but the degradation is now visible, matching `discover_configs`' own `Warning: skipping ...` convention:

```python
except Exception as exc:  # noqa: BLE001 - SessionStart must never hard-fail
    # Broad on purpose: a discovery bug must not block the session. But the
    # fallback below sees one config at most, so staying silent here is the
    # very under-installation this function exists to prevent - say so.
    print(f"Warning: eval.yaml discovery failed ({exc}); falling back to a "
          f"single config. .eval-venv may be missing deps required by other "
          f"evals in this project.", file=sys.stderr)
```

What a user sees:

```
Warning: eval.yaml discovery failed (import of agent_eval.config halted; None in sys.modules);
falling back to a single config. .eval-venv may be missing deps required by other evals in this project.
```

Added two tests: one asserting the warning fires on discovery failure, one asserting it does *not* fire on the happy path. `tests/test_ensure_deps.py` is now 16 tests; full suite `869 passed, 2 skipped`.

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

PR #184, opened 2026-08-13
feat(judges): honor score_range in the LLM judge request and enforce it everywhere

### LINKED ISSUE #182: LLM judges always use a 1-5 score schema, ignoring score_range
## Summary

For LLM judges, `score_range` is never applied. The judge is asked for an integer 1–5 no matter what the config says, and the returned value is not clamped or validated against the configured range. A judge declared as

```yaml
- name: testability_score
  prompt_file: eval/prompts/testability-judge.md
  feedback_type: int
  score_range: [0, 2]
  samples: 3
```

can return `4`, and `4` is what lands in `summary.yaml`.

The agent judge path honors `score_range` correctly. Only the LLM path ignores it, which makes the inconsistency easy to miss.

## Where

`skills/eval-run/scripts/score.py` at `main` (v1.37.2, 7b63484).

The tool schema is a module-level constant with the range baked in:

```python
_SCORE_SYSTEM_PROMPT = (
    "You are a judge evaluating skill outputs. Call the submit_score tool "
    "once with an integer score 1-5 and a thorough rationale.")

_SCORE_JUDGE_TOOL = {
    "name": "submit_score",
    ...
            "score": {"type": "integer", "minimum": 1, "maximum": 5,
                      "description": "Overall score, 1 (worst) to 5 (best)."},
}
```

`_call_structured_judge(prompt, model, feedback_type, images=None, max_tokens=4096)` selects that tool for anything that is not `bool`, and its signature never receives the `JudgeConfig`. The LLM scorer closure has `jc` in scope but does not pass it:

```python
feedback_type = "bool" if jc.feedback_type == "bool" else "score"

def scorer(outputs=None, **kwargs):
    ...
    return _call_structured_judge(rendered, judge_model, feedback_type,
                                  images=images)
```

So `jc.score_range` reaches neither the schema, nor the system prompt, nor any post-hoc clamp.

Contrast the agent judge path, which does both. It tells the judge the range:

```python
elif jc.score_range:
    verdict_spec = ('{"score": <number in [%s, %s]>, '
                    '"rationale": "<short justification>"}'
                    % (jc.score_range[0], jc.score_range[1]))
```

and clamps the result in `_interpret_agent_verdict`:

```python
if jc.score_range:
    lo, hi = jc.score_range
    value = max(lo, min(hi, value))
```

## Observed behavior

In one run, three LLM judges with identical config (`feedback_type: int`, `score_range: [0, 2]`, `samples: 3`) scoring the same artifact against a 0–2 rubric:

| judge | recorded value | scale actually used |
|---|---|---|
| feasibility | 1 | 0–2 |
| scope | 1 | 0–2 |
| testability | 4 | 1–5 |

The testability judge's own rationale opened with "Testability score: **1**." — it had read the 0–2 rubric out of the prompt file and scored on it — and then submitted `4` through the tool, because the tool is the only channel and the tool says 1–5. All three samples returned 4, so the run looked perfectly stable.

## Why this is worth fixing rather than working around

- **It is silent.** No error, no warning, no validation failure. The value is in range for the schema, just not for the rubric.
- **Per-sample stability cannot detect it.** The judge is consistently wrong, so `stable: true` and the min/max/mean all agree.
- **Scores become non-comparable across judges.** A mean over dimensions on different scales, and any `min_mean` threshold applied to it, is not measuring what the config says it is.
- **The bottom of the range is unreachable.** With `"minimum": 1`, a judge configured for `[0, 2]` can never return the 0 band, so the rubric's "fails this dimension" outcome cannot be expressed.
- **Prompt-side workarounds do not help.** The rubric in the prompt file already said 0–2, and the judge acknowledged it in the rationale. The tool schema overrides the prompt.

## Suggested fix

Thread the range into the call and enforce it on the way out:

1. Pass `jc.score_range` (or `jc`) into `_call_structured_judge`.
2. Build the score tool per call rather than using a module-level constant, setting `minimum`/`maximum` and the `description` from the range, and defaulting to 1–5 when `score_range` is absent so current behavior is unchanged.
3. Derive the system prompt's "integer score 1-5" text from the same range.
4. Clamp after parsing, matching `_interpret_agent_verdict`, so the text-fallback path is covered too.

An optional extra: warn when a judge returns a value outside `score_range` rather than clamping silently, since a clamp can mask a prompt/rubric mismatch.

Happy to send a PR if the approach looks right.

danmcp:
Fixes #182.

`score.py` built the `submit_score` tool and its system prompt from module-level constants with `1-5` baked in, so a judge declaring `score_range: [0, 2]` was asked for a 1-5 score and nothing checked what came back.

@astefanutti's analysis on the issue reframed the defect, and this PR is built around that reframing: the bug is the **missing invariant, not the fixed scale**. `minimum`/`maximum` on a non-strict `input_schema` are advisory, so the 44 out-of-range zeros in the strat-creator corpus arrived by exactly the same unchecked route as the reported `4`. Rebuilding the tool schema alone would fix nothing reliably. So the scale is now stated where the model will act on it (system prompt + schema), and enforced centrally on the way out.

`agent_eval/harbor/reward.py:196-202` and `agent_eval/evalhub/adapter.py` both load `score.py` by path, so this one change covers local, Harbor and EvalHub.

## Shape

A `_numeric_bounds(jc) -> (lo, hi, is_int)` helper feeds three call sites, as suggested: the per-call LLM tool + system prompt, the agent `verdict_spec`, and validation. `_call_structured_judge` keeps its primitive signature with a trailing `bounds=None` kwarg — `feedback_type` stays third positional, so the existing positional callers in `tests/test_score_builtin.py` are untouched.

## Point by point

**1. Fail loudly on out-of-range; don't clamp.** `_enforce_bounds` raises `ScoreRangeError` and the sample is recorded as an error, so it drops out of the aggregate instead of being imputed. Clamping was the actively harmful option: a `4` from a `[0, 2]` judge clamps to `2` — a perfect score that lifts the mean and paints the cell green.

The breach is **persisted**, not stderr-only: it lands on the judge result → `summary.yaml` → the report, the same path every judge error already takes. The stderr `WARNING` is an addition on top of persistence, not the record — and it is deliberately the *only* judge error that prints, since a scale breach is a config/prompt bug that recurs every run, while printing every error would just add noise to a `ThreadPoolExecutor` pass.

`report.py:_score_band_class` now returns `fail` for a value off the scale rather than banding it green.

**2. Reward normalization, bundled.** `compose_reward` takes a `judge_ranges` mapping and normalizes each numeric judge over its **own** declared `score_range`, falling back to `score_min`/`score_max` only for judges that declare none. A `judge_ranges(config)` helper builds it, and `skills/eval-anova/scripts/analyze.py` passes it through `load_conditions_from_runs`, so the ANOVA/Pareto dependent variable is no longer flattened for evals without a `reward:` block. On a `[0, 2]` judge the old default mapped 0 and 1 both to 0.0 and the top band to 0.25.

**3. The text fallback is range-aware and raises.** `_parse_score_response` takes `bounds` and derives its patterns from them: `(\d)\s*/\s*5` becomes `/<hi>`, and the last-resort scan keeps only numbers that are **on the judge's scale** instead of `\b([1-5])\b`. The `3` default is gone — unparseable now raises so the sample is recorded as an error, matching the agent judge. Any default here (the literal `3`, or a scale midpoint) is a fabricated score that counts toward the mean.

**4. `feedback_type` typing.** `is_int` comes from `feedback_type`, selecting `"integer"` vs `"number"` in the schema; parsing goes through `float()` and rounds only for `int`. A `_fmt_bound` helper renders `2` rather than `2.0`, since config load coerces `score_range` to floats and `"0.0-2.0"` in the prompt invites fractional scores.

**5. Scale-shift note.** In `README.md`, under the `feedback_type`/`score_range` key-concepts entry — a short "Upgrading" callout saying that judges with a non-default range were graded on `[1, 5]` regardless of what they declared, that their means shift, and to re-baseline `thresholds.min_mean` before treating a drop as a regression. It is in the README rather than `CHANGELOG.md` because `.releaserc.json` generates that file with `@semantic-release/release-notes-generator` on the angular preset, which emits commit **subjects** only and rewrites the file on release. No `BREAKING CHANGE:` footer, per your read.

### Smaller things

- **`_make_builtin_scorer` hard-codes `"bool"`** — now enforced at config load: a builtin **LLM** judge declaring `feedback_type` or `score_range` is rejected rather than having the declaration silently dropped at scoring time. Builtin *Python* judges are unaffected; they may be numeric.
- **Config-load validation**, following the `reward.judge` / `judge.step` convention: `feedback_type: bool` with a `score_range` is rejected, `feedback_type: int` with fractional bounds is rejected, and a numeric LLM/agent judge that omits `score_range` warns.
- **`_judge_score_ranges` `int()` truncation** — bounds keep their fractional part, reconciled with `config.py`'s float coercion. `[0, 2.5]` no longer becomes `(0, 2)` and bands a legitimate 2.4 as off-scale. Only the histogram needs whole numbers and it derives its own bins with `floor`/`ceil`.
- **MLflow `make_judge` fallback** — scoped out, stating it explicitly as you asked. It still receives `feedback_value_type` but not the range. Values from that path are not unguarded, though: enforcement sits at the shared `scorer()` call site, so an off-scale result is still recorded as an error sample — the judge simply isn't *told* the scale.
- **`score_min`/`score_max` Jinja vars** — not included, taking the "optional" at face value. Easy follow-up if wanted.

## Additions to the list

A few things that seemed to follow from the analysis rather than sit beside it:

- **Unknown `builtin:` names fail at load.** A typo like `cost_budgets` used to survive until scoring, after the run had been paid for. Resolution uses a filesystem scan (`builtin_judge_kind` / `builtin_judge_names`) rather than `BuiltinJudgeRegistry.discover()`, which execs every Python judge module — far too much work to validate a config. Names are matched against the scan rather than used to build a path, so a `builtin:` value cannot address anything outside the package.
- **Agent judges are validated, not clamped.** `_interpret_agent_verdict`'s `max(lo, min(hi, value))` is gone; enforcement is central, so an agent judge that ignores its scale errors exactly like an LLM judge. This makes the two paths agree in the direction the issue argued for, rather than agreeing on clamping.
- **The histogram no longer hides off-scale readings.** `_ascii_score_hist` widened its bins to cover values outside the declared range — your point that `range(lo, hi+1)` drops the very sample worth seeing applied whether or not the value is later rejected.
- **The numeric-without-range warning exempts inline `check:` judges**, which compute their own value, so a raw count stays legal. For the same reason, only a **declared** `score_range` is enforced: a judge that declares none keeps emitting whatever it emits rather than being failed against the `[1, 5]` default.

## Tests

59 new tests. The centrepiece is `tests/test_score_builtin.py::TestJudgeRequestPayload`, which asserts on the request itself — `messages.create.call_args.kwargs["system"]` containing `0-2` and not `1-5`, and `kwargs["tools"][0]["input_schema"]["properties"]["score"]` carrying `minimum`/`maximum` from the range and `"integer"` vs `"number"` from `feedback_type`, plus the `[1, 5]` default when nothing is declared. Nothing asserted on the payload before, which is why this shipped.

Around it: `tests/test_score_range_enforcement.py` (the error path end to end, and that a breach is persisted), `TestParseScoreResponseBounds` (prose fractions, the loose scan ignoring off-scale numbers, unparseable raising, `float` keeping its decimal), `tests/test_config_score_scale.py` (24 tests over the new load-time rules and builtin resolution), `tests/test_report_score_bands.py` (fractional bounds surviving, off-scale banding as fail, histogram widening) and `tests/test_harbor_reward.py` (per-judge normalization). `tests/test_agent_judge.py::test_score_range_clamps` became the error-sample assertion it now describes.

The four commits are independently green and bisectable. The full suite passes apart from 7 failures that predate the branch and are unrelated (Vertex client construction and responses-API lifecycle).

## Validation

Beyond the suite, this ran end to end in the strat-creator eval's CI: four judges declared `feedback_type: int`, `score_range: [0, 2]`, sampled 3×, all returning values on their declared scale (1/1/0/1) with no off-scale readings, no error samples and no `ScoreRangeError`. That eval is what surfaced the bug.

Consumer-side companion PR: opendatahub-io/strat-creator#59 — a defect in that eval's own grounding gate, found while validating this fix end to end.

## Summary by CodeRabbit

* **New Features**
  * Added configurable integer and fractional score ranges for numeric judges.
  * Added score-range enforcement across prompts, reports, agent verdicts, and rewards.
  * Added per-judge reward normalization and optional error-rate thresholds.
  * Reports preserve fractional scores and identify out-of-range results.

* **Bug Fixes**
  * Out-of-range scores now become errors instead of being clamped or fabricated.
  * Invalid judge configurations and unknown built-in judges are rejected with clearer validation.

* **Documentation**
  * Expanded guidance for score ranges, verdict types, reward scaling, errors, and configuration rules.


astefanutti:
Thanks — this is a good fix, and the reframing landed exactly right: treating the missing invariant as the defect rather than the fixed scale is what makes it hold up.

I've reviewed it and pushed 12 commits to the branch rather than sending you a list to work through. They're all on top of `910db49`, so your four commits are untouched. Happy to drop or rework any of them — two in particular are judgement calls I'd rather you made (below).

Everything below was reproduced before and after. Suite goes 945 → 971 passing, each commit independently green.

## First, a correction to my own review

I claimed the report and `anova.json` would now disagree because `report.py` wasn't threaded with `judge_ranges`. **That was wrong.** `_render_reward_overview` only renders when the config has a `reward:` block (`report.py:2902`), and in that case `reward_cfg` is non-None, so `compose_reward` short-circuits to `compute_reward_from_config` before reading `judge_ranges` at all:

```
HEAD, reward_cfg (production)  -> 0.2500
HEAD, reward_cfg=None (tests)  -> 1.0000
main, reward_cfg (production)  -> 0.2500   (unchanged)
```

The divergence I described cannot arise. `11bf31e` still threads it — the three call sites should agree — but its message now says plainly that this is consistency rather than a live bug, and names the gap that *is* live: `compute_reward_from_config` normalizes every judge through `reward.score_range` and ignores per-judge ranges. I deliberately left that alone; it's pre-existing documented behaviour and changing it belongs in its own PR, not inside a score_range fix.

## The two that mattered

**`b71f442` — the line that fixes #182 had no test.** Deleting `bounds=bounds` from `_load_llm_judge`, reverting the reported bug outright so a `[0, 2]` judge is told "integer score 1-5" again, left the suite at **944 passed / 12 failed — byte-identical to unmutated**. `TestJudgeRequestPayload` hands `_call_structured_judge` its bounds directly, which covers the schema builder and nothing upstream. Three new tests go through `_load_llm_judge` from a `JudgeConfig` and assert the captured request; two of them now fail under that mutation.

**`a851591` — `_enforce_bounds` was rewriting values.** It ended in `_coerce_number(value, is_int)`, and `is_int` is true for any `feedback_type` other than `"float"` — including the unset default, which `config.py:574` documents as "Inferred if omitted" but never infers. `judge_bounds` is built for every judge with a `score_range`, so this reached judges that never went near a model:

```yaml
- {name: ratio, score_range: [0, 1], check: "return (0.75, 'r')"}
# before: value 0.75    after: value 1    (no warning, no error)
```

Enforcement is now validate-only. Coercion stays in the three paths that turn a model's answer into a number, all of which already did it — so LLM and agent judges are unaffected.

Related, `df4b16d`: `score_range: [0, 2.5]` with no `feedback_type` was accepted at load and produced `"an integer score 0-2.5"` with schema `{"type": "integer", "minimum": 0, "maximum": 2.5}` — an unreachable maximum, and exactly the combination `config.py` rejects when spelled `feedback_type: int`. It also defeated `4dcd321`, which taught the report to band a scale the scorer couldn't produce a value on. When `feedback_type` is absent the intent now comes off the scale: whole bounds mean a banded rubric, fractional bounds a continuous one. `[0, 2]` and the `[1, 5]` default are unchanged.

## Two calls I'd like you to make

**`19e9595` changes pre-existing tested behaviour.** `compose_reward` returns 1.0 when no numeric judge produced a value — right for a gates-only config, wrong for a trial whose scorers all failed, and range enforcement turns "all failed" from a rare exception into the normal consequence of a model ignoring its rubric:

```
q=1, r=1  on [0, 2]  -> 0.5    (obeys the scale)
q=4, r=4  on [0, 2]  -> 1.0    (breaches it, both ScoreRangeError)
```

Breach rate correlates with whatever factor a sweep is testing, so the inversion lands on the independent variable. `report.py`'s `if not metrics` guard doesn't catch it either — one passing boolean judge makes `metrics` non-empty. But `test_compose_reward_ignores_skipped_and_errored` asserted the old result on `main`, so this is a contract change, not a regression fix. The commit says "drop this commit if you disagree" and is cleanly removable.

**`df636d9` is noisy at scale.** The new warning gated on `feedback_type in ("int", "float")`, but the default is `""` and `_numeric_bounds` treats anything non-`bool` as numeric — so the judge that actually runs on the unenforced default, one declaring neither field, was the one shape it could never reach. Adding `""` fixes that but emits **89 warnings across 26 real configs**, largely one per judge per Harbor task config, i.e. inside every trial container. All true positives. You may prefer it once per load.

## Smaller ones

- `0e21ef6` — `_load_agent_judge` hand-rolled `_numeric_bounds`' rule and had already drifted: a judge with no declared range was told `{"score": <number>}`, no scale at all. Contract construction unified; `_interpret_agent_verdict`'s looser rounding left alone and commented, since it has a test of its own.
- `191dbd4` — the `make_judge` fallback was the one path where a judge is failed against a scale it was never given. `make_judge` takes no scale argument, but the instructions are the same channel the Anthropic path's system prompt uses.
- `7db4105` — `num` had no sign, so on a `[-1, 1]` judge every prose fallback read the verdict backwards (`"Overall score: -1"` → `1`), which `_enforce_bounds` can't catch because the flipped value is in range.
- `57a96ff` — `int()` ran before the widening, so an off-scale 2.9 on a `[0, 2]` judge became an in-range 2: the reading the widening exists to expose was the one it hid. `_MAX_HIST_BINS` was also inert — widening happened first and the cap's fallback was that same span, so `_ascii_score_hist(2, [2, 2, 100000], 0, 2)` built a 100,010-character glyph. Also asserts the `ScoreRangeError` stderr WARNING; disabling the print had left the suite green.
- `dcb6c4e` — three doc corrections, including your `reference/config/judges.md` builtin-LLM rule (CodeRabbit is right that `bool` is permitted) and `concepts/reward-api.md`, which stopped being true once the default composition started normalizing per judge.

## CodeRabbit triage

| Comment | Verdict |
|---|---|
| Non-finite bounds (`.nan`/`.inf`) | **Not a regression** — `[0, .inf]` raises on both `main` and this branch, just at different points; `.nan` is dropped on both. Downgrade to a nit. |
| Histogram `int()` truncation | **Valid**, fixed in `57a96ff`. |
| MD046 lint | **Invalid** — markdownlint isn't wired into CI, and the MD046 count is identical on `main` (3, at unrelated lines). Rewrote the passage anyway. |
| builtin-LLM `bool` wording | **Valid**, fixed in `dcb6c4e`. |
| `capsys` stderr assertion | **Valid**, added in `57a96ff`. |

## Method note

The last commit, `0b4132e`, is fixes from an adversarial pass I ran over my own 11 commits before pushing. It found seven defects, two of them real: the histogram marker landed on an empty bin for any fractional median, and `0e21ef6`'s test asserted `_numeric_bounds` rather than the `verdict_spec` that commit actually changed — so reverting it left the suite byte-identical. That is precisely the failure `b71f442` exists to call out, reproduced by me three commits later. Both kill their mutants now. Two commit messages also stated things that turned out to be false and have been reworded.

Remaining gaps worth knowing about: nothing pins the three `compose_reward` consumers to one another, so a fourth call site would be invisible; and `compose_reward(reward_cfg=..., judge_ranges=...)` together is untested — `reward_cfg` silently wins.

## Release

All commits are `fix:`/`docs:`/`refactor:`, so semantic-release cuts a **patch**. For a change that converts previously-scored values into error samples, shifts reward magnitude, and adds hard config-load rejections, I'd rather this went out as a minor — happy to retitle the core commit `feat:` on merge if you agree.

astefanutti:
Pushed five more commits: the three CodeRabbit findings, plus a documentation sweep that turned up more than I expected.

## CodeRabbit

All three were valid. Reproduced each before fixing.

**`073d793` — NaN bypassed the bounds check.** Correct and the only silent hole: NaN compares False against everything, so `value < lo or value > hi` waved it through and one sample turned the judge's whole mean into NaN. `inf` was already caught (`inf > hi` is true).

```
_enforce_bounds(float("nan"), (0, 2, True), "j")  ->  nan
```

**`c1265d7` — agent verdicts now round on the scale their contract states.** This is the one I'd noted and deliberately left; CodeRabbit reaching the same conclusion independently changed my mind. `_interpret_agent_verdict` rounded on an explicit `feedback_type: int`, while the contract handed to the agent says "integer" for any whole-numbered scale — so one config scored differently depending on the runner:

```
score_range: [0, 5], no feedback_type, verdict 3.5
  LLM path   -> 4
  agent path -> 3.5
```

**This changes pre-existing behaviour** — `test_numeric_float_preserved_without_int_type` asserted the 3.5 on `main`. Its intent (a float survives without `feedback_type: float`) still holds and is still tested, now on a fractional scale where the contract asks for a number; a second case pins the whole-scale rounding. Second such commit in the series, same caveat: droppable on its own.

**`5c04606` — `raw`/`normalize` exceptions.** My earlier rewrite overcorrected. `reward.score_range` does not apply to `raw` judges, and in single-judge mode only when `normalize: true`.

## Documentation

I audited every surface mentioning `score_range` rather than just the pages this PR already touched, and the docs were describing three different contracts, none of them current.

**`154129d` is the one to look at first, and it isn't docs.** `validate_eval.py`'s judge-field allowlist had drifted from `JudgeConfig`, missing `score_range`, `step` and `agent`. The repo's own example fails today:

```
$ validate_eval.py config eval/anova-example/eval.yaml
ERROR: judges.solution_quality has unknown field(s): score_range
INVALID: anova-example
```

Pre-existing, but blocking here: everything this PR writes tells authors to declare `score_range`, and `/eval-analyze --validate` then calls the result invalid. Pinned both ways against the dataclass, matching the `STANDARD_TEMPLATE_VARS` contract test already in that file — a list that had drifted twice before, which is why the pin exists.

**`eb1956f` — 32 files.** The two that matter most were generating the problem rather than describing it: `eval-yaml-template.md` is what the agent copies when writing a user's `eval.yaml`, and it framed `score_range` as report colouring; `analyze-skill.md` actively instructed the agent that "omitting the field is the right default". Every generated config inherited both.

The rest, grouped:

- **A `[0, 1]` default that never existed** — `_numeric_bounds` falls back to `[1, 5]` for every non-bool judge, and an undeclared judge renders *neutral*, not banded.
- **"Inferred if omitted" on `feedback_type`** — nothing infers it, and it is now load-bearing.
- **Default reward composition** described as a flat `[1, 5]` normalize in four places, with the code block predating the all-errored branch.
- **Every shipped numeric-LLM-judge example warned at load** — including the root `eval.yaml`, which tells you to copy it into your project, so a new user's first command warned. 20 examples fixed; a sweep now reports zero. Binary rubrics got `feedback_type: bool` rather than a fabricated 1-5 scale.

Also added where a user would actually look: enforcement applies to `check`/`module`/builtin Python judges too (in the cookbook, beside the judges it affects — that is a real behaviour change for anyone who added a range for colouring); the verbatim load warning and scoring-time stderr line; the unknown-`builtin:` load error; what an error sample does to a mean, a threshold and the report; and an upgrade note on the judges page, since the README is not part of the published site.

Two corrections I made to my own earlier commits while doing this: the claim that a config with an explicit `reward:` block is unaffected was wrong — its reward moves too, because the declared scale now reaches the model, so the judge answers on a different scale before `reward.score_range` is applied. Only the *normalization basis* change is default-path-only. That was in both `README.md` and `concepts/judges.md`.

## State

975 passed, 12 pre-existing environment failures, unchanged throughout. Root `eval.yaml` loads clean under `-W error::UserWarning`. Every fenced YAML example containing `judges:` across all touched files was extracted and run through `EvalConfig.from_yaml` — no errors, no warnings.

Still open, deliberately: `compute_reward_from_config` normalizes every judge through `reward.score_range` and ignores per-judge ranges. It's pre-existing, documented, and now documented *accurately*, but it's the remaining inconsistency in this area and probably wants its own PR.

danmcp:
### End-to-end validation run

Quick mode, one case (`RHAIRFE-1056`), `claude-opus-4-6`. The eval reports FAIL; both breaches are explained below and neither comes from this change.

**What it validates:** all four rubric judges declare `feedback_type: int`, `score_range: [0, 2]`, and every one returned a value on that scale — 1 / 1 / 1 / 2. No off-scale values, no `ScoreRangeError`, no error samples, all stable across 3 samples. Before this change they were asked for a 1-5 score and nothing checked the answer.

**The two breaches:**

- **`scope_score` 1.0 vs `min_mean: 1.5`** — the floor was tuned against means produced while these judges were being answered on `[1, 5]`. On the declared `[0, 2]` scale it asks for better than "adequate" on every case. Re-baselining is tracked separately and needs a full multi-case sweep.
- **`pipeline_total` 4.0 vs `min_mean: 5.0`** — this judge does **not** sum the four rubric dimensions. It reads `scores.total` from the review frontmatter — the pipeline's own assess-strat score — and exists to measure pipeline-vs-independent-judge agreement. The pipeline scored itself 4/8 while the independent judges total 5/8; that 1-point gap is the metric working. The 4 is a real pipeline result, unaffected by this PR.

**Caveat:** N=1. The previous run of this same case scored scope 0 and architecture 1 where this one scores 1 and 2, so nothing here is a basis for setting thresholds.

<details>
<summary>Full run summary</summary>

**Eval:** strat-creator-eval | **Model:** claude-opus-4-6 | **Run ID:** 2026-08-13-opus-4-6 | **Result:** FAIL

| Judge | Type | Result | Threshold | Status |
|---|---|---|---|---|
| architecture_context_used | check | pass_rate: 1.0 | >= 0.7 | PASS |
| architecture_score | agent | mean: 2.0 | >= 1.0 | PASS |
| cost_budget | builtin | pass_rate: 1.0 | >= 0.8 | PASS |
| feasibility_score | llm | mean: 1.0 | >= 1.0 | PASS |
| files_exist | check | pass_rate: 1.0 | >= 1.0 | PASS |
| frontmatter_valid | check | pass_rate: 1.0 | >= 1.0 | PASS |
| pipeline_total | check | mean: 4.0 | >= 5.0 | **FAIL** |
| recommendation_consistency | check | pass_rate: 1.0 | >= 1.0 | PASS |
| scope_score | llm | mean: 1.0 | >= 1.5 | **FAIL** |
| strategy_nonempty | check | pass_rate: 1.0 | >= 1.0 | PASS |
| testability_score | llm | mean: 1.0 | >= 1.0 | PASS |

| Case | Feasibility | Testability | Scope | Architecture | Pipeline Total |
|---|---|---|---|---|---|
| RHAIRFE-1056 | 1 | 1 | 1 | 2 | 4 |

| Metric | Value |
|---|---|
| Cases | 1 |
| Duration | 1048.6 s |
| Cost | $5.69 |
| Turns | 97 |
| Cache hit rate | 91.18% |
| Cost per turn | $0.05866 |
| Cost per Mtok | $1.2669 |
| Output tokens per turn | 499.27 |

</details>

Note: This was before your additions. Rerunning now.

danmcp:
Both of the calls you flagged look good to me — keep `19e9595` and `df636d9` as they are.

Agreed on the release too: please retitle the core commit `feat:` on merge.

Thanks for pushing the fixes as well. The `_enforce_bounds` value rewriting and the untested `bounds=bounds` wiring were both real, and I'd have missed the `validate_eval.py` allowlist drift entirely.

astefanutti:
Addressed all 10 of the latest round. Two were real gaps rather than nits, and both are in the threshold path — reachable precisely because this branch turns off-scale values into error samples.

**The diagnostic blamed the wrong thing.** A judge whose every value was rejected reported *"mean unavailable — judge skipped for all cases or not a numeric judge"*. It wasn't skipped; it errored, and the rationales say why. `score_cases` now counts errored cases per judge so the two causes can be distinguished:

> mean unavailable — judge errored on 3 cases; see the per-case rationales

A genuinely skipped judge keeps the old wording.

**`min_mean` gated a shrinking sample** — the review's Major, and it's right. Errors don't lower a mean, they remove cases from it, so:

```
{"mean": 2.0, "values": [2.0], "errored_cases": 9}
  min_mean: 1.0                       -> passes, exit 0
  min_mean: 1.0, max_error_rate: 0.2  -> error_rate 0.900 (9 of 10 cases errored)
```

I added `max_error_rate` as a fourth threshold key rather than failing on any partial error. Turning one flaky judge run into a suite failure would be worse than the hole it closes; declaring the key is how CI states how much of the dataset actually has to be scored. Off unless declared, so no existing config changes behaviour.

The other eight, all verified before fixing:

- **`validate_eval.py`** — the unknown-field diagnostic still listed the old allowlist. Now generated from `valid_judge_fields`, which is the drift the allowlist fix was about in the first place. Good catch.
- **The allowlist contract test** kept `condition` in its expected set, so an accidental `condition` entry would have passed the test written to prevent exactly that. Now exact equality against the dataclass minus `condition` plus `if`.
- **`builtin-judges.md`** documented Python builtins as returning `(passed: bool, …)` two lines above a note saying they may be numeric.
- **`report.md`** grouped boolean judges under numeric score-range banding.
- **`reward.md`** — the `1.0` fallback also covers numeric judges skipped by their `if:` condition, not just configs with no numeric judges.
- **The SkyRL doc** described averaging as unconditional.
- **`CLAUDE.md`** stray slash, and two unused `rat` bindings (RUF059 — advisory here, CI runs skillsaw not ruff; I renamed only the two that are genuinely unused, four others do assert on the rationale).

980 passing, same 12 pre-existing environment failures.

One I'd push back on gently if you disagree: `max_error_rate` is new config surface on a PR that was meant to be a fix. It follows the existing `min_*` key pattern and is inert unless declared, so I think it earns its place — but it's the last commit and drops cleanly.

Stacked #185 has been rebased onto this head; one conflict in the SkyRL doc where both branches edited the same paragraph, resolved to keep both facts.

astefanutti:
Thanks for the validation run and the calls — I'll retitle the core commit `feat:` on merge, and I'm leaving `19e9595` and `df636d9` as they are.

The N=1 caveat is the right read. Worth noting the two breaches are the two thresholds with essentially no margin in the archive: `scope_score` sat at 1.5417 against its 1.5 floor across the earlier runs, and two of them landed on exactly 1.5. So a single case moving is enough either way — which is a re-baselining problem, not a signal.

One thing your run does settle: all four judges returned values on the declared `[0, 2]` scale with no error samples, which is the first end-to-end confirmation that the enforcement path doesn't reject well-behaved judges.

Addressed the one new comment, plus three more instances a sweep turned up. `max_error_rate` is a maximum, but the prose around it still said *"Every threshold is a minimum: the run passes when the metric is `>=` the value"* — self-contradicting on the same page. Fixed in the concepts intro, its heading and key table, the reference intro and comparison note, the unknown-key admonition (*"only the three keys above are honored"*) and the `guides/ci.md` gate table. The last two weren't in the review.

Also stated the exact-equality rule once instead of per key: the three `min_*` keys compare with `<` and `max_error_rate` with `>`, so a metric equal to its threshold passes either way.

980 passing, same 12 pre-existing environment failures. #185 rebased onto this head.

danmcp:
### Re-ran end to end against the full branch

This covers your 17 commits, not just the original four. Quick mode, one case (`RHAIRFE-1056`), `claude-opus-4-6`. Same two threshold breaches as before, neither from this PR.

**What it validates:** all four rubric judges declare `feedback_type: int`, `score_range: [0, 2]`, and every one returned a value on that scale, stable across 3 samples. No off-scale values, no `ScoreRangeError`, no error samples.

Two things worth calling out that the earlier run didn't exercise:

- **`scope_score` returned 0** — the bottom of the declared range, accepted rather than rejected. The previous run had no judge land on a boundary, so this is the first confirmation that enforcement isn't over-firing at the edge of the scale.
- **`df636d9` stayed silent.** The load produced no `score_range` warnings at all, because this eval declares a range on every numeric judge. That's the shape the warning is meant to leave alone, and it does.

**The two breaches**, both unchanged from the last run:

- **`scope_score` 0.0 vs `min_mean: 1.5`** — the floor was tuned against means produced while these judges were answered on `[1, 5]`. Re-baselining is tracked separately and needs a full multi-case sweep.
- **`pipeline_total` 4.0 vs `min_mean: 5.0`** — this judge doesn't sum the four rubric dimensions. It reads `scores.total` from the review frontmatter — the pipeline's own assess-strat score — and exists to measure pipeline-vs-independent-judge agreement. Unaffected by this PR.

**On the variance:** scores moved against the previous run on the same case — scope 1 → 0, architecture 2 → 1. Across three observations of `RHAIRFE-1056` scope has now come in at 0, 1, 0. This reinforces the earlier caveat rather than undercutting it: N=1 is not a basis for setting thresholds, which is exactly why the re-baselining wants a multi-case sweep.

<details>
<summary>Full run summary</summary>

**Eval:** strat-creator-eval | **Model:** claude-opus-4-6 | **Run ID:** 2026-08-13-opus-4-6 | **Result:** FAIL

| Judge | Type | Result | Threshold | Status |
|---|---|---|---|---|
| architecture_context_used | check | pass_rate: 1.0 | >= 0.7 | PASS |
| architecture_score | agent | mean: 1.0 | >= 1.0 | PASS |
| cost_budget | builtin | pass_rate: 1.0 | >= 0.8 | PASS |
| feasibility_score | llm | mean: 1.0 | >= 1.0 | PASS |
| files_exist | check | pass_rate: 1.0 | >= 1.0 | PASS |
| frontmatter_valid | check | pass_rate: 1.0 | >= 1.0 | PASS |
| pipeline_total | check | mean: 4.0 | >= 5.0 | **FAIL** |
| recommendation_consistency | check | pass_rate: 1.0 | >= 1.0 | PASS |
| scope_score | llm | mean: 0.0 | >= 1.5 | **FAIL** |
| strategy_nonempty | check | pass_rate: 1.0 | >= 1.0 | PASS |
| testability_score | llm | mean: 1.0 | >= 1.0 | PASS |

| Case | Feasibility | Testability | Scope | Architecture | Pipeline Total |
|---|---|---|---|---|---|
| RHAIRFE-1056 | 1 | 1 | 0 | 1 | 4 |

| Metric | Value | Previous run |
|---|---|---|
| Cases | 1 | 1 |
| Duration | 1009 s | 1048.6 s |
| Cost | $3.82 | $5.69 |
| Turns | 76 | 97 |
| Cache hit rate | 89.47% | 91.18% |
| Cost per turn | $0.0503 | $0.05866 |
| Cost per Mtok | $1.5587 | $1.2669 |
| Output tokens per turn | 609.63 | 499.27 |

</details>

danmcp:
Keep max_error_rate. It reads as new surface, but it's the gap this PR opens rather than something unrelated riding along.

Agreed on the re-baselining read too. scope_score has now come in at 0, 1, 0 across three observations of the same case, which is more evidence that the 1.5 floor was set against the old [1, 5] answers and needs a multi-case sweep.

astefanutti:
Answering "is anything left in the docs" turned up a **code bug in `max_error_rate`** — my own, from two commits ago. Fixed in `2644397`, along with the doc gaps.

## The bug

The gate derived its denominator from `values`, which `cmd_judges` strips before writing `summary.yaml`:

```python
name: {k: v for k, v in agg.items() if k != "values"}
```

So on the standalone path — `score.py regression --run-id …`, which `guides/ci.md` documents as *the* CI gate — `scored` came back 0 and every judge with a single error read as 100% errored:

```
live     {"values": [4.0]*9, "errored_cases": 1}  -> no regression
summary  {"errored_cases": 1}                     -> error_rate 1.000
```

Any adopter of the new key would have got a red build at any threshold below 1.0. My test only exercised the in-memory shape, so the suite stayed green — which is the same gap `b71f442` was added to close, reintroduced by me one commit later. `scored_cases` is now persisted alongside `errored_cases`, with a `len(values)` fallback for summaries written before this, and both shapes are tested.

## Docs

`max_error_rate` was missing from every enumeration of the valid keys. Two of them — `CLAUDE.md:220` and `README.md:359` — describe theirs as exhaustive, and `CLAUDE.md` is the standing context for agents editing `eval.yaml` here, so that one was actively misleading. Also the glossary, the `eval-yaml` schema map, four nav lists, the `skill-case` table, and the concepts flowchart, which branched on three keys under a "four threshold keys" heading and routed the new one through a `<` comparison it doesn't use.

Two more worth calling out:

- **`skills/eval-analyze/references/eval-yaml-template.md`** now offers the gate. That template is what puts an enforced `score_range` on every generated judge, so without it every generated config came out enforced-but-ungated.
- **`reference/config/thresholds.md`** still said the regression detail "blames a skip" — behaviour `_unavailable_reason` removed earlier in this same PR. It contradicted its sibling page, which already quoted the new string.

`runs-directory.md` also gains the two new `summary.yaml` fields; its key list said `aggregated` where the file actually writes `judges`.

## State

982 passing, same 12 pre-existing environment failures. #185 rebased onto this head — one conflict in `CLAUDE.md` where both branches rewrote adjacent bullets, resolved to keep both.

One follow-up I'm **not** doing here, since it predates this PR: `report.py` and `log_results.py` only know `min_pass_rate`/`min_mean`, so a `max_error_rate` breach fails CI while `report.html` shows no Regressions row and MLflow tags `regressions_detected=no`. `min_win_rate` has the identical gap on `main` today, so it's a standing divergence between the scoring CLI and its two consumers rather than something this PR introduces — worth its own issue covering both keys.

astefanutti (approved):
Thanks @danmcp!

PR #185, opened 2026-08-13
fix(reward): normalize each composed judge over its own score_range

astefanutti:
Makes the configured reward path honor each judge's own `score_range`, closing the last place a declared scale is ignored.

> [!IMPORTANT]
> **Stacked on #184.** The first four commits here are that PR; review from `3d16542` onward. GitHub will show the combined diff until #184 merges, after which this reduces to 17 files.

## The bug

`compute_reward_from_config` — the path taken whenever `eval.yaml` has a `reward:` block — normalized every numeric judge through the single global `reward.score_range`. One pair of numbers for N judges cannot be right for more than one of them:

```
testability [0, 2] = 2   (a perfect score)
clarity     [1, 5] = 4
equal weights, reward.score_range: [1, 5]  ->  0.500
                                  per-judge ->  0.875
```

Worse for RL, a short scale loses most of its gradient. On `[0, 2]` normalized against `[1, 5]`, values 0 and 1 both map to 0.0 and 2 maps to 0.25 — two of three rubric bands collapse into one reward, which is the entire signal a GRPO loop learns from. Now 0.0 / 0.5 / 1.0.

#184 fixed this for `compose_reward`'s default path and threaded `judge_ranges` into all three production callers, but `compose_reward` dropped it the moment a `reward:` block existed. So the two paths disagreed: **adding an equal-weight `reward:` block silently changed the number**, 0.875 → 0.500. It no longer does.

## Precedence

First rule that applies:

1. boolean → `1.0` / `0.0`, no range consulted (`gate` runs before everything);
2. judge in `reward.raw`, or the single `reward.judge` **without** `normalize: true` → clamped to `[0, 1]`, no range consulted;
3. the judge's own declared `score_range`;
4. `reward.score_range` — deprecated fallback;
5. `[1, 5]`.

The judge's own range wins because it is already authoritative everywhere else: it goes into the judge's prompt and tool schema, the answer is enforced against it, and the report bands the cell with it. Normalizing that number against a different range is a unit error.

One shared `_resolve_range` now serves both paths, so they cannot drift apart again.

## `reward.score_range` is deprecated, not removed

It parses to `None` when absent (`effective_score_range` resolves it), which is what makes the deprecation warning targetable. It fires only where the precedence change actually moves a number:

```
reward.score_range [1.0, 5.0] is deprecated and no longer normalizes
'testability' [0.0, 2.0]: a judge's own 'score_range' wins. It still
applies to 'plain'; drop it once every composed judge declares a 'score_range'.
```

Silent when the key is absent, when the ranges agree, when the conflicting judge is in `raw`, when it is not composed at all, and for a clamped single judge — none of which consult a range.

**Deliberately not added:** a migration warning for configs that omit the field. It would fire on correctly written new configs and need an env var to silence; the README upgrade note carries that instead.

## Compatibility

Rewards move for any config with a `reward:` block whose judges declare their own ranges. No config in this repo has a `reward:` block — only documentation examples, all updated — but external RL users should **re-baseline rather than mix pre- and post-upgrade rollouts**. That instruction is in the README upgrade note and in `3d16542`'s body.

`RewardConfig.score_range` becoming `Optional[list]` is a real API change for anything doing `lo, hi = cfg.reward.score_range`; `effective_score_range` lands in the same commit as the migration.

## `355e745` is separable

A different pre-existing footgun, surfaced by this work and **droppable on its own** — it changes no numbers. `reward: {judge: x}` clamps rather than normalizes, so on a scale wider than `[0, 1]` every value at or above 1 is a perfect reward, silently. Now warns unless the judge declares `[0, 1]`, declares nothing, or sets `normalize: true`.

## Testing

**1007 passing**, 12 pre-existing environment failures (missing `anova` extra, responses-API integration), unchanged throughout. Each commit is independently green.

32 new tests. **No existing test needed editing** — `judge_ranges` is keyword-only, and that was the acceptance gate for the compatibility claim: if an existing test had needed a change, the claim was wrong.

The one worth pointing at is `test_compose_reward_threads_ranges_into_the_configured_path`, which spies on the call and asserts `judge_ranges` arrives. #184 demonstrated that wiring a range mapping into two of three call sites is exactly how this fails, and that single line is the whole fix.

Every `reward:` example in the docs was extracted and loaded through `EvalConfig.from_yaml` — no errors, no warnings.

## Follow-ups, not in scope

- `reward.weights` / `reward.raw` names are not validated against defined judges; only `reward.judge` is.
- `reward.judge_ranges(config)` and `report.py::_judge_score_ranges(config)` are two extractors that must agree — this change makes more behaviour depend on that.
- Composing a judge on a deliberately different scale than it reports on is no longer expressible via `reward.score_range`; the replacement is an expression over already-normalized names.

## Summary by CodeRabbit

* **New Features**
  * Reward calculations now normalize numeric judge scores using each judge’s declared score range across all reward compositions.
  * Added clearer handling for raw values, single-judge rewards, weighted rewards, and formula-based rewards.
* **Bug Fixes**
  * Improved consistency between configured and default reward calculations.
  * Added warnings for deprecated or conflicting reward-level score ranges and unnormalized single-judge scores.
* **Documentation**
  * Updated configuration, upgrade, reporting, and reward guidance to explain normalization, clamping, fallback behavior, and reinforcement-learning implications.


astefanutti:
All three addressed. Two were code, and the first was worse than described.

**Non-finite bounds — `0d90461`.** Both validators leaked, each a different value, because they spell the same check two different ways:

```
judge   `if lo >= hi`        -> NaN fails every comparison, so it passes
reward  `if not lo < hi`     -> catches NaN, but `0 < inf` is True
```

So `[nan, 2]`, `[0, nan]`, `[0, inf]` and `[-inf, 0]` were all accepted at judge level, and the infinite pair at reward level too. The consequences are silent rather than loud: `_normalize` over an infinite span returns `0.0` for every score, collapsing the reward signal, and the judge is asked for *"a numeric score 0-inf"* in its system prompt. Both now require finite increasing bounds via `math.isfinite`.

Worth noting this is the same class as the `.nan`/`.inf` comment on #184, which I triaged down to a nit at the time. That was right for what it described — a crash in the histogram, present on `main` too — but it was pointing at a real hole one layer up that I didn't follow. Good catch on the re-raise.

**The clamp warning — same commit.** Correct, and the fix generalises further than suggested. `normalize: true` is needed for *any* non-`[0, 1]` scale, but the two failure modes are opposite: `[0, 2]` saturates (everything ≥ 1 is a perfect reward) while `[0, 0.5]` can never reach the top of the reward range. The warning now names whichever applies:

```
[0, 2]   -> …clamped to [0, 1] — every score at or above 1 becomes the maximum reward.
[0, 0.5] -> …clamped to [0, 1] — the reward can never exceed 0.5.
```

**Docs — `b3c2172`.** The precedence statement was written as if every numeric judge is normalized over its own range, omitting the two cases that consult no range at all. Fixed in seven places, and the unconditional "`[1, 5]` fallback" replaced with the real chain where it appeared. `reference/config/reward.md` keeps the canonical five-rule section; the others carry a short clause and a link rather than restating it.

Four of the locations in the list were already correct — `reward-api.md` L93-105, `reward-rl.md`, `README.md` L329 and `CLAUDE.md` L221 all carried the exceptions — so those are untouched.

1025 passing, same 12 pre-existing environment failures.

astefanutti:
Follow-up on the saturation comment — I'd only half-fixed it.

The previous commit corrected the *code* warning but left the prose copies. `analyze-skill.md` still said clamping "saturates — every score at or above 1 becomes the maximum reward", which is the exact claim you flagged as wrong for a narrow scale. Sweeping for it turned up three more: the reward reference, the RL cookbook, and the eval.yaml template.

All four now name whichever direction applies — a scale reaching 1 saturates, a narrower one caps the reward below the top. `analyze-skill.md` matters most of the four since it steers what `/eval-analyze` generates.

`63275c5`. 1025 passing.

The other open thread (`config.py` non-finite bounds) is addressed in `0d90461` — `math.isfinite` on both validators — GitHub just hasn't marked the thread outdated since the line is still nearby.

PR #186, opened 2026-08-13
fix(report,mlflow): agree with the CLI on what counts as a regression

astefanutti:
`score.py::detect_regressions` is the gate CI exits on. `report.py` and `log_results.py` each reimplemented a subset of its rules, and both had drifted to two of the four threshold keys — so a run could exit 1 while `report.html` rendered no Regressions table and MLflow tagged `regressions_detected=no`.

Follow-up to #184, where this surfaced. `min_win_rate` has had the gap since it was introduced; `max_error_rate` inherited it on arrival.

## What was wrong

Three implementations of the same rules:

| | `min_mean` | `min_pass_rate` | `min_win_rate` | `max_error_rate` |
|---|---|---|---|---|
| `score.py` (exit code) | ✅ | ✅ | ✅ | ✅ |
| `report.py` Regressions table | ✅ | ✅ | ❌ | ❌ |
| `report.py` per-judge column | ✅ | ✅ | ❌ | ❌ |
| `log_results.py` MLflow tag | ✅ | ✅ | ❌ | ❌ |

The per-judge column was the worst of the three: a judge gated **only** on `max_error_rate` rendered a green `PASS` while the run exited 1.

## What changed

Rather than adding two more branches in three places, both consumers call the detector:

- `_render_regressions` builds its table from it.
- `_render_scoring_summary` still picks one bound to display in the Threshold column — four bounds in one cell helps nobody — but takes the PASS/FAIL verdict from the detector.
- `log_results.py` loads `score.py` by path, the same way `agent_eval/harbor/reward.py` already loads the judge engine. It degrades to "no regressions" if the load fails, so MLflow logging never breaks on this.

## Testing

`tests/test_threshold_consumers.py` runs one table of cases — one per key, plus a clean run — through all three surfaces and asserts they agree. Reverting `report.py` to its old two-key form fails three of them.

998 passing, 12 pre-existing environment failures (missing `anova` extra, responses-API integration), unchanged.

## Note

Independent of #185: this branches off `main` and touches no file that PR touches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized regression detection across command-line checks, reports, and MLflow results.
  * Reports now show every breached evaluation threshold, including coverage-only failures.
  * PASS/FAIL statuses and regression tags now consistently reflect all configured thresholds.

* **Tests**
  * Added coverage to verify consistent regression results across CLI output, reports, and MLflow tagging.

### DIFF
--- a/skills/eval-mlflow/scripts/log_results.py
+++ b/skills/eval-mlflow/scripts/log_results.py
@@ -50,22 +50,20 @@ def _detect_regressions(judges, thresholds):
     """score.py's regression detector, loaded by path.
 
     The judge engine lives with the eval-run skill rather than in the
-    agent_eval package — same approach as agent_eval/harbor/reward.py. Returns
-    [] if it can't be loaded, so MLflow logging never fails on this.
+    agent_eval package — same approach as agent_eval/harbor/reward.py. Raises
+    if it cannot be loaded; the caller tags the run "unknown" rather than
+    claiming a clean one.
     """
     import importlib.util
-    try:
-        root = Path(__file__).resolve().parent.parent.parent.parent
-        score_path = root / "skills" / "eval-run" / "scripts" / "score.py"
-        if not score_path.exists():
-            return []
-        spec = importlib.util.spec_from_file_location("_score_for_thresholds",
-                                                      score_path)
-        mod = importlib.util.module_from_spec(spec)
-        spec.loader.exec_module(mod)
-        return mod.detect_regressions(judges, thresholds)
-    except Exception:
-        return []
+    root = Path(__file__).resolve().parent.parent.parent.parent
+    score_path = root / "skills" / "eval-run" / "scripts" / "score.py"
+    if not score_path.exists():
+        raise FileNotFoundError(f"Judge engine not found: {score_path}")
+    spec = importlib.util.spec_from_file_location("_score_for_thresholds",
+                                                  score_path)
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+    return mod.detect_regressions(judges, thresholds)
 
 
 def _resolve_skill(config: EvalConfig) -> str | None:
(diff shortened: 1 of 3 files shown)


astefanutti:
@coderabbitai review

astefanutti:
Both valid, both fixed in `159c0f8`. The first is broader than reported.

## The status branch ran before the verdict

`if pass_rate is None and mean is None` came first and settled on SKIP/ERROR, so the threshold branch below it never ran for a judge whose gated metric isn't a column in this table.

You flagged `min_win_rate`. It also hit `max_error_rate` when **every** case errored — and my own test missed that, because it used a partially-errored judge that still had a mean and so went down the working branch:

```
win_rate 0.3 vs min_win_rate 0.6          report SKIP  / CI fails
all 10 cases errored, max_error_rate 0.2  report SKIP  / CI fails
min_mean with no surviving case           report SKIP  / CI fails
```

Rather than reordering around the fallback, the breach check now runs **first** — it's the authoritative one, and the metric behind it need not be displayable. Falling out of that: a *satisfied* threshold now reads PASS instead of SKIP beside a populated Threshold column, which was the same bug wearing the other face.

## The detector's own error handling was the bug this PR is about

You're right, and it's worse stated plainly: both helpers caught everything and returned `[]`, so a failure to load `score.py` was indistinguishable from finding no regressions. The report rendered no Regressions section, the summary rendered PASS, and MLflow tagged `regressions_detected=no` — while CI evaluated the same thresholds and could fail the run.

That's precisely the silent-wrongness this PR removes, reintroduced in its own error handling. I wrote that fallback deliberately ("degrades rather than failing the logging step") and it was the wrong call.

Both helpers now raise:

- the report renders `Could not evaluate thresholds: <reason>` instead of an empty section, and no longer claims PASS on a verdict it never obtained (`_breach_known` gates it);
- `regressions_detected` gains a third value — `unknown`. `no` now means the detector ran and found nothing.

Tests: the parametrised table covers all five status cases, and reverting the breach-first reorder fails two of them. 1006 passing, same 12 pre-existing environment failures.

PR #204, opened 2026-08-26
feat(workspace): symlink-free {dest, source} shared files, portable across substrates

astefanutti:
## Description

Follow-up to the review on #203. Adds a **symlink-free `{dest, source}` shared-file** form to `dataset.workspace.files`, and makes it **portable across all three execution substrates** (local, Harbor, S3/EvalHub).

Today, referencing a live SKILL.md from a case means committing a *symlink* into the case dir. That is fragile: it is skipped on the local path, **ignored entirely by Harbor** (`tasks.py` only stages `input.yaml`/`answers.yaml`), and does not round-trip through object storage (a symlink object downloads as a text file, and an out-of-case target can't even be a per-case S3 object). This PR replaces that pattern with an explicit, materialized entry.

```yaml
dataset:
  workspace:
    files:
      - src/                         # existing: per-case path (string) — unchanged
      - dest: triage-skill.md        # new: shared, symlink-free
        source: skills/address-ci-failures/SKILL.md
```

`source` is resolved at packaging time against the project root or a configured `runner.plugin_dirs` entry (symlinks followed, containment-checked), then **copied — never symlinked** — into every case at `dest`. Because the result is a real file, the same entry works everywhere:

| Substrate | How the shared file gets there |
| --- | --- |
| **Local `/eval-run`** | Materialized into each per-case workspace (`workspace.py` → `materialize_shared_files`). |
| **Harbor** | Materialized from the local checkout into each task package's `environment/` (both single-step and multi-step blocks), so it travels inside the package — no ConfigMap/bind-mount/image needed. |
| **S3 / EvalHub** | Materialized into each per-case dir by a new export step (`skills/eval-dataset/scripts/export_s3.py`) **before** upload, since a pod loaded from S3 has no project/plugin files to resolve `source` against at run time. `download_dataset` then restores it per case with no adapter change. |

### Design

- One shared implementation backs all three paths: `config.resolve_workspace_source` (the trust-boundary resolver) + `agent_eval.workspace_provisioning.materialize_shared_files`. The trust boundary and copy semantics can't drift between substrates.
- A `source` that is missing, dangling, or resolves outside the project/plugin roots is **skipped with a warning** (never pulls a host file in — CWE-59). Directory sources copy recursively with nested symlinks dropped.
- A `dest` colliding with a harness-reserved name (`input.yaml`, `answers.yaml`, `batch.yaml`, `.claude`, …) is skipped with a warning, so a shared file can't clobber harness-managed files or expose the answer key.
- Fully backward compatible: string entries are parsed and copied exactly as before.

### Files

- `agent_eval/config.py` — `WorkspaceFile` dataclass, `resolve_workspace_source`, loader parsing for `{dest, source}`.
- `agent_eval/workspace_provisioning.py` (new) — shared materializer.
- `skills/eval-run/scripts/{workspace.py,workspace_files.py,report.py}` — local wiring + dict-aware report.
- `agent_eval/harbor/tasks.py` — materialize into `environment/` (both blocks).
- `agent_eval/evalhub/export.py` (new) + `skills/eval-dataset/scripts/export_s3.py` (new) — S3/EvalHub export step.
- Docs: config reference, CLAUDE.md, eval-dataset SKILL.md, Harbor + EvalHub READMEs.

## How Has This Been Tested?

- 20 new unit tests: config parsing (accept/reject dict entries), local materialization (in-project source, followed source symlink, plugin-dir source, out-of-bounds skip, missing skip, reserved-dest skip, nested-symlink drop), Harbor wiring (shared across cases + out-of-bounds skip), and S3 export (materialize per case, drop case symlinks, out-of-bounds skip, follow source symlink, optional S3 upload key layout).
- Full suite green locally: `python3 -m pytest tests/` → **1408 passed, 10 skipped**, incl. `test_venv_activation` (the new `export_s3.py` CLI carries the `agent_eval._bootstrap` first-import).

## Notes

- Branch is based on `main` and is **orthogonal to #203**: this handles dict `{dest, source}` entries; #203 handles string-symlink entries. They compose (minor conflict at most on `workspace_files.py`).
- Harbor coverage: the single-step block has a direct wiring test; the multi-step block calls the identical `materialize_shared_files` helper (unit-tested).

## Merge criteria

- [x] Commits squashed with meaningful messages.
- [x] Testing instructions in the PR body.
- [x] Manually verified the changes work (full suite green).

## Summary by CodeRabbit

- **New Features**
  - Added shared workspace file configuration using destination/source mappings.
  - Shared files are securely copied into local, Harbor, and exported evaluation workspaces.
  - Added dataset export to self-contained case directories, with optional S3 uploads.
  - Added support for exporting directories while safely handling symlinks.

- **Bug Fixes**
  - Improved validation for unsafe, missing, reserved, or unauthorized workspace sources.

- **Documentation**
  - Updated configuration, dataset publishing, Harbor, and EvalHub guidance.


astefanutti:
Rebased onto `main` (which now includes the merged #203 — listed **string** file-symlink materialization; this PR adds the complementary symlink-free **`{dest, source}`** form). Addressed all CodeRabbit findings:

- **🟠 Major — `dest` accepts `"."` (config.py):** Good catch — `Path(".").parts == ()` bypassed the reserved-name guard and `dirs_exist_ok=True` would copy a directory source over the workspace root. Fixed by passing `reject_root=True` to `_validate_relative_path` for `dest`, so `.` is rejected at config load. Added `test_workspace_files_dict_dot_dest_rejected`.
- **🟡 Minor — reserve `annotations.yaml` (workspace_provisioning.py):** Added to `DEFAULT_RESERVED_NAMES` so a shared file can't clobber case metadata (it's exposed in the EvalHub export / whole-case copy).
- **🟡 Minor — raw-string regex in tests:** The re-appended parsing tests now use raw-string `match=` patterns (`r"..."`).
- **📘 Doc — field-table type (dataset.md):** Updated the `workspace.files` row to `list of strings or {dest, source} maps` and documented both entry forms.
- **🔵 Nitpick — B904 in export_s3.py:** `raise SystemExit(1) from None` in the boto3 `ImportError` handler.

Full suite green locally: **1419 passed, 9 skipped**.

astefanutti:
Thanks — both new findings were valid; fixed in the amended commit:

- **🟠 `dest: "/"` bypassed the root guard (config.py + workspace_provisioning.py).** Real hole: `"/".rstrip("/") == ""`, and `_validate_relative_path("")` returns early on the empty string, so the empty dest slipped past `reject_root` and (via `Path("").parts == ()`) past the reserved-name check → `dst == target_dir`. Fixed at load with `dest.rstrip("/") or "."` (so `reject_root` catches it), **and** added defense-in-depth in `materialize_shared_files` — a dest that is empty/root, absolute, or contains `..` is now skipped with a warning (covers programmatically built `WorkspaceFile`s that bypass config validation). Added `test_workspace_files_dict_slash_dest_rejected` + `test_materialize_shared_file_skips_empty_dest`.
- **📘 committed-symlink statement (dataset.md).** Corrected — since #203 merged, a committed *file* symlink to an approved-root target is now materialized locally; it's still ignored by Harbor and doesn't round-trip through S3, and directory/out-of-root symlinks are skipped everywhere. The `{dest, source}` form remains the portable one.

Full suite: **1421 passed, 9 skipped**.

astefanutti:
@coderabbitai review

astefanutti:
Addressed the latest round (and one I'd missed from the prior round):

- **🟠 Major — recursive source→dest copy (workspace_provisioning.py).** Real DoS: `source: "."` with a target dir under the project root would `copytree` the destination into itself (CWE-400). Now I resolve `dst` and skip with a warning when it equals the source or resolves inside a directory source. Test: `test_materialize_shared_file_skips_recursive_dest`.
- **🟡 `RuntimeError` on a symlink-loop `plugin_dir` (config.py `workspace_source_roots`).** The `except` caught `(ValueError, OSError, TypeError)` but not `RuntimeError` from `Path.resolve()` on a loop. Added `RuntimeError` so one bad plugin dir is dropped, not fatal. Test: `test_materialize_shared_file_tolerates_symlink_loop_plugin_dir`.
- **🟡 dataset.md — scope the per-case rule.** The "companion files must exist in every case" note now states it applies to **string** entries only; a shared `{dest, source}` needs no per-case file.
- **🟡 dataset.md — document rejected `dest` forms.** The table and prose now spell out that empty/`.`/`/`, absolute, and parent-traversing dests are rejected at load and skipped at materialization, and added a row for a dest resolving inside the source.

Full suite **1423 passed, 9 skipped**; `mkdocs --strict` clean.

PR #205, opened 2026-08-27
fix(config): skip hidden files and dirs in eval-config discovery

astefanutti:
pathlib's `glob("*.yaml")` matches dotfiles — unlike the `glob` module — so a hidden working file under `eval/` surfaces as an eval config. Observed concretely: a project keeping a git-ignored `eval/.entity-map.yaml` (an anonymization map, deliberately hidden and never a config) gets it discovered as an eval named `.entity-map`, and since discovery then reports two configs, `/eval-run`'s single-config auto-selection turns into a which-config prompt on every invocation.

Hidden files and directories are never configs under any documented layout (root / `eval/<name>.yaml` flat / `eval/<name>/eval.yaml` nested), so both scans now skip dotfile entries. Regression test covers a hidden flat file and a hidden nested dir alongside a real flat config.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hidden files and directories in the evaluation configuration folder are no longer detected as evaluation configurations.
  * Prevents temporary or working files from appearing in configuration discovery results.

* **Tests**
  * Added coverage verifying that hidden configuration files and directories are ignored.

### DIFF
--- a/agent_eval/config.py
+++ b/agent_eval/config.py
@@ -1622,12 +1622,17 @@ def discover_configs(project_root: Path) -> list[DiscoveryResult]:
     eval_dir = project_root / "eval"
     if eval_dir.is_dir():
         for subdir in sorted(eval_dir.iterdir()):
-            if subdir.is_dir():
+            if subdir.is_dir() and not subdir.name.startswith("."):
                 candidate = subdir / "eval.yaml"
                 if candidate.is_file():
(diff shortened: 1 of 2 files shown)


astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

PR #209, opened 2026-08-27
fix(staging): copy plugin-root .mcp.json into isolated eval workspaces

smg247:
## Summary
- Isolated evals stage a plugin into `.staged-plugins/` and pass that path to Claude. Staging copied the manifest, skill roots, and `commands`/`agents`/`hooks`/`scripts`, but not plugin-root `.mcp.json`.
- Claude Code loads plugin MCP servers from that file, so skills that depend on them (for example SHIP Status in `plugins/ci`) ran without those tools in the catalog.
- `stage_plugin_dir` now copies `.mcp.json` with the same symlink-escape rules as the optional dirs.

## Test plan
- [x] `python -m pytest tests/test_stage_plugins.py` (22 passed)
- [ ] Confirm a payload-analysis eval init event includes `mcp__ship-status__*` / `mcp__plugin_ci_ship-status__*` after this lands

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Plugin staging now includes a valid root-level `.mcp.json` configuration file.
  * Staged plugin MCP configuration is available during runner execution.

* **Bug Fixes**
  * Invalid, dangling, escaping, or non-file `.mcp.json` links are safely excluded from staging.
  * Existing symlink containment, atomic-copy, and repeatable staging behavior remain supported.

* **Documentation**
  * Updated staging documentation to describe `.mcp.json` handling.

### DIFF
--- a/agent_eval/agent/claude_code.py
+++ b/agent_eval/agent/claude_code.py
@@ -26,8 +26,8 @@ _print_lock = threading.Lock()
 # ${CLAUDE_PLUGIN_ROOT}/scripts/... or ${CLAUDE_SKILL_DIR}/../../scripts/...
 # at runtime; a staged copy without it would break those plugins.
 _PLUGIN_OPTIONAL_DIRS = ("commands", "agents", "hooks", "scripts")
-# Plugin-root MCP config (Claude Code loads this from the plugin directory).
-# Staging omitted it, so isolated evals never saw plugin MCP servers.
+# Optional plugin-root files Claude Code loads besides the manifest and
+# conventional directories. Copied only when present.
 _PLUGIN_OPTIONAL_FILES = (".mcp.json",)
 
 # Bulk plugin discovery never reads — keeps the staged copy small.
--- a/agent_eval/agent/claude_code.py
+++ b/agent_eval/agent/claude_code.py
@@ -79,6 +79,10 @@ def stage_plugin_dir(plugin_dir: Path, workspace: Path) -> Path:
     manifest (via ``resolve_plugin_skill_roots``, which validates
     containment), the conventional ``_PLUGIN_OPTIONAL_DIRS`` when they
     exist at the plugin root, and ``_PLUGIN_OPTIONAL_FILES`` (``.mcp.json``).
+    Isolated MCP coverage is that conventional config file plus servers
+    already reachable without extra trees (HTTP/npx/uvx, or code under
+    ``scripts/``). Local stdio trees (``servers/``, …) and a custom
+    ``mcpServers`` path in the manifest are not staged.
     Symlinks are not reproduced — in-plugin targets are copied as content,
     dangling ones are skipped, and links escaping the plugin are refused —
     so the staged tree cannot point back outside the workspace. Idempotent
--- a/agent_eval/agent/claude_code.py
+++ b/agent_eval/agent/claude_code.py
@@ -26,6 +26,9 @@ _print_lock = threading.Lock()
 # ${CLAUDE_PLUGIN_ROOT}/scripts/... or ${CLAUDE_SKILL_DIR}/../../scripts/...
 # at runtime; a staged copy without it would break those plugins.
 _PLUGIN_OPTIONAL_DIRS = ("commands", "agents", "hooks", "scripts")
+# Plugin-root MCP config (Claude Code loads this from the plugin directory).
+# Staging omitted it, so isolated evals never saw plugin MCP servers.
+_PLUGIN_OPTIONAL_FILES = (".mcp.json",)
 
 # Bulk plugin discovery never reads — keeps the staged copy small.
 _PLUGIN_IGNORE = shutil.ignore_patterns(".git", "node_modules", "__pycache__")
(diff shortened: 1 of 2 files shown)


agent_eval/agent/claude_code.py
```diff
@@ -137,6 +140,17 @@ def stage_plugin_dir(plugin_dir: Path, workspace: Path) -> Path:
...
+            source = plugin / name
+            if not source.is_file():
+                continue
+            try:
+                resolved = source.resolve(strict=True)
+            except (OSError, RuntimeError):
+                continue
+            if not resolved.is_file() or not resolved.is_relative_to(plugin):
+                continue
+            shutil.copy2(resolved, staging / name)
```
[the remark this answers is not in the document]
smg247: The resolve-then-reopen gap is pre-existing on the optional-dir `copytree` path just above, which already names CWE-367 and copies from the resolved pathname for the same reason. This loop matches that mitigation. Closing it properly (openat / `O_NOFOLLOW` / copy from the fd) belongs on both roots, not only `.mcp.json` — leaving that as a follow-up rather than a one-sided fix here.

astefanutti, line 157: **[minor] The referenced MCP server _code_ isn't staged — only the config file.**

This copies `.mcp.json` verbatim but nothing it points at. A stdio server declared as `{"command": "node", "args": ["${CLAUDE_PLUGIN_ROOT}/servers/index.js"]}` will get a staged `.mcp.json` whose `${CLAUDE_PLUGIN_ROOT}` resolves to the staged path — but `servers/` (or `mcp/`, `dist/`, `bin/`, …) isn't in `_PLUGIN_OPTIONAL_DIRS`, so it was never staged. The server then fails to launch and the skill still runs without its tools — the exact symptom this PR targets.

The plugins-reference's own idiomatic example uses `${CLAUDE_PLUGIN_ROOT}/servers/db-server` + `${CLAUDE_PLUGIN_ROOT}/config.json` + `${CLAUDE_PLUGIN_ROOT}/data`, none of which are staged.

Not a regression (previously nothing loaded), and it doesn't affect npx/uvx external servers or code under the already-staged `scripts/`. Your unchecked test-plan box — confirm the init event includes `mcp__ship-status__*` — is exactly the check that would catch this for the real target plugin; worth completing before relying on it. At minimum, a one-line docstring note that only external-command or already-staged-dir servers are covered under isolation would set expectations.

agent_eval/agent/claude_code.py
```diff
@@ -26,6 +26,9 @@
 # ${CLAUDE_PLUGIN_ROOT}/scripts/... or ${CLAUDE_SKILL_DIR}/../../scripts/...
 # at runtime; a staged copy without it would break those plugins.
 _PLUGIN_OPTIONAL_DIRS = ("commands", "agents", "hooks", "scripts")
+# Optional plugin-root files Claude Code loads besides the manifest and
+# conventional directories. Copied only when present.
+_PLUGIN_OPTIONAL_FILES = (".mcp.json",)
```
astefanutti, line 31: **[minor] Manifest-declared custom `mcpServers` path isn't staged (hardcoded `.mcp.json` only).**

Per the plugins-reference, `plugin.json`'s `mcpServers` may be a `string`/`array` path (e.g. `"./mcp-config.json"`), not just the conventional root `.mcp.json`. A manifest pointing at a custom filename gets neither the file nor its contents staged, so the skill runs in the isolated eval without those MCP tools — the same silent tool-loss this staging exists to prevent.

This is the same convention-vs-manifest gap the repo _deliberately_ solved for skills via `resolve_plugin_skill_roots` (config.py). Inline-object `mcpServers` already rides along inside the staged `plugin.json`, so only the custom-path variant is affected.

Follow-up (not this PR): read `manifest.get("mcpServers")` and, when it's a string/array path, stage those entries with the same containment check, keeping the `.mcp.json` fallback — mirroring `resolve_plugin_skill_roots`.

astefanutti (commented):
Two non-blocking follow-up notes on the `.mcp.json` staging (inline). The fix itself is correct and the premise checks out against the plugin docs (`.mcp.json` at the plugin root is the default auto-loaded MCP location) — these are completeness gaps, not merge blockers.

agent_eval/agent/claude_code.py
smg247: Agreed — this PR only stages the conventional config, not a local stdio tree. HTTP/npx/uvx and code already under `scripts/` still work; `servers/` and friends would be a follow-up (and would widen what isolation copies).

We did complete that check locally: a one-shot Claude against the staged `plugins/ci` tree showed `plugin:ci:ship-status` connected with `mcp__plugin_ci_ship-status__*` in the init catalog. Noted the coverage limit on `stage_plugin_dir`.

agent_eval/agent/claude_code.py
smg247: Leaving the custom-path case as a follow-up, same as `resolve_plugin_skill_roots` for skills. Conventional `.mcp.json` is what this eval needs. Documented on `stage_plugin_dir` so the gap is explicit.

astefanutti (approved):
Thanks @smg247!

PR #213, opened 2026-08-31
feat: add Cursor as an eval runner

Prashanth684:
Add first-class Cursor Agent CLI support to the eval harness:

- Register and execute `runner.type: cursor` with model effort, plugins, environment, timeouts, usage reporting, and stream-json output.
- Translate common permissions to Cursor's temporary `.cursor/cli.json`, restoring existing project configuration after execution.
- Support Cursor skill loading and isolated/repo workspace execution.
- Normalize Cursor tool events for conversation, cost, and documentation tracking.
- Route synthetic generation and non-Anthropic LLM judges through the configured runner with a strict JSON verdict contract.
- Update validation, CLI, EvalHub/Harbor handling, documentation, and tests.

Known limitations:

- Cursor does not consume the generated Claude `.claude/settings.json` in repo mode and uses `--force`, so those auto-generated repository Write/Edit denies are not enforced. Explicit `permissions.allow`/`deny` rules are translated to Cursor where possible, but unsupported mappings and `inputs.tools` remain limitations. Prompt guidance and post-run checks are not hard prevention.
- Cursor does not enforce `max_budget_usd`; provider-side limits apply.

Verification: focused runner, event, prompt-backend, configuration, synthetic generation, and judge tests pass; syntax and diff checks pass.

## Summary by CodeRabbit

- **New Features**
  - Added support for running evaluations with Cursor Agent locally.
  - Added Cursor-specific permission handling, effort settings, plugin skills, workspace modes, and environment configuration.
  - Added runner-backed prompt generation and LLM judging for supported non-Anthropic models.
  - Added support for per-case and shared dataset workspace files.

- **Bug Fixes**
  - Improved event, conversation, and consulted-file tracking for Cursor execution.
  - Added validation with clearer errors for unsupported runner settings.

- **Documentation**
  - Documented Cursor configuration, permissions, authentication, limitations, and runner-specific behavior.
  - Clarified that Cursor is unavailable on Harbor and EvalHub.


skills/eval-run/scripts/score.py
```diff
@@ -2072,48 +2215,51 @@ def _load_llm_judge(jc, config, project_root=None):
...
+    use_anthropic = bool(anthropic_auth and is_anthropic_model(judge_model))
 
-        def scorer(outputs=None, **kwargs):
-            out = outputs or {}
-            rendered = _render_jinja2_template(prompt, arguments, out)
-            images = _extract_images(out)
+    # Preserve the existing MLflow/OpenAI-compatible fallback for Claude
+    # models when direct Anthropic credentials are absent. Non-Anthropic model
+    # ids are runner-managed and must use the model-agnostic path below.
+    if not use_anthropic and is_anthropic_model(judge_model):
```
astefanutti: **Regression (high).** This gate makes `make_judge` reachable only for Claude models. A real non-Anthropic judge (e.g. `gpt-4o` + `OPENAI_API_KEY`) can no longer reach it: `use_anthropic` is False *and* `is_anthropic_model` is False, so this branch is skipped and it falls through to `_call_structured_judge_via_runner`, which runs the agent CLI with `--model gpt-4o` (e.g. `claude --model gpt-4o`) and errors on every case. A config that worked before this PR now breaks.

**Minimal fix for this PR:** keep `make_judge` as the non-Anthropic path (its gateway already serves OpenAI/Anthropic/Gemini/Azure/Bedrock/…) and make the runner-backed judge path an explicit opt-in rather than the blanket default for everything non-Anthropic. Full provider-URI routing → follow-up PR.

skills/eval-run/scripts/score.py
```diff
@@ -2072,48 +2215,51 @@ def _load_llm_judge(jc, config, project_root=None):
...
+        try:
+            from mlflow.genai.judges import make_judge
+        except ImportError:
+            pass
+        else:
+            instructions = prompt
+            if bounds:
```
astefanutti: **Regression (medium).** Gating the scale instruction on `if bounds:` (always truthy for a numeric judge) emits "return a score between 1 and 5; a score outside that range is rejected, not clamped" even for numeric judges that declared no `score_range`. But enforcement only covers *declared*-range judges — `judge_bounds` is built as `{jc.name: _numeric_bounds(jc) for jc in config.judges if jc.score_range}` (~score.py:1552), so an undeclared-range judge is never range-checked. Net effect: the instruction is untrue (nothing rejects out-of-range) **and** it needlessly compresses a judge that may have been intended for e.g. 0–100 into 1–5.

**Fix:** revert to `if jc.score_range:` so the stated scale matches what's actually enforced (stated ⟺ enforced).

agent_eval/agent/cursor_agent.py
```diff
@@ -0,0 +1,1026 @@
...
+            config = json.loads(original) if original else {}
+            if not isinstance(config, dict):
+                raise ValueError(f"Cursor config must be a JSON object: {config_path}")
+            permissions = config.get("permissions") or {}
+            if not isinstance(permissions, dict):
+                raise ValueError("Cursor config permissions must be a JSON object")
+
+            if allow_rules:
+                allow_patterns = _cursor_permission_patterns(allow_rules, strict=True)
+                permissions["allow"] = allow_patterns
```
astefanutti, line 921: **Bug (high).** If `allow_patterns` is empty — e.g. an allow list whose tools have no Cursor mapping — this writes `permissions["allow"] = []`, which per the surrounding logic means *deny-everything*. The agent the user tried to **grant** access to ends up able to do nothing, silently and with no error.

Suggest: raise in strict mode when a non-empty allow list maps to zero Cursor patterns, or fall back to the all-permissions default instead of writing an empty allow list.

astefanutti (commented):
**Review: Cursor runner** — the Cursor-specific code is carefully done (nice symlink/atomic-write hardening in the permission handling). Findings split into two buckets:

- **Fix in this PR** (bugs in the new code + regressions this PR introduces to existing judge routing): #1–#6, plus #7 as a nit.
- **Separate follow-up PR** (larger capability, out of scope here): the full "judge provider decoupled from runner" design — let a judge model be a `provider:/model` URI, plumb `model=` into `make_judge` (today it's dropped, so that path is pinned to `openai:/gpt-4.1-mini`), a `runner:/…` explicit opt-in for Cursor-internal models, `generate_synthetic.py` parity, and adding `litellm` to the venv for the long tail. We'll do that separately.

On **#1** specifically: the minimal in-PR fix is to keep `make_judge` reachable for non-Anthropic judges (pre-PR behavior) and make the runner-backed judge path an explicit opt-in — *not* to force every non-Anthropic model through the runner. The full provider-URI routing lands in the follow-up.

Inline comments below.

agent_eval/events.py
```diff
@@ -804,7 +1109,15 @@ def extract_conversation_text(events, include_thinking=False):
...
         text = event.get("text", "")
         if text:
-            parts.append(text)
+            if event.get("_cursor_delta"):
+                saw_cursor = True
+                cursor_deltas.append(text)
+            else:
+                parts.append(text)
+    if saw_cursor:
+        cursor_text = cursor_result if cursor_result is not None else "".join(cursor_deltas)
```
astefanutti: **Bug (medium).** When a Cursor `result` event carries `result: ""`, `_cursor_result_text` is set to `""` (a str, so `cursor_result is not None`), then `cursor_text = ""`, the `if cursor_text:` guard below is false, and all accumulated assistant deltas are dropped — the conversation surfaces as empty despite real content.

Suggest falling back to the deltas when the result text is empty, e.g. `cursor_result or "".join(cursor_deltas)` (or treat an empty result as absent).

agent_eval/prompt_backends.py
```diff
@@ -0,0 +1,192 @@
...
+from typing import Optional
+
+from agent_eval.agent import RUNNERS
+from agent_eval.events import extract_conversation_text, parse_stream_events
+
+
+_ANTHROPIC_MODEL_ALIASES = {"opus", "sonnet", "haiku"}
+
+
+def is_anthropic_model(model: Optional[str]) -> bool:
```
astefanutti, line 25: **Bug (medium).** `is_anthropic_model` matches only the exact aliases `{opus, sonnet, haiku}`, a `claude` substring, or an `anthropic/` prefix. Versioned/bracketed Claude ids that lack the literal "claude" — e.g. `sonnet-4-5`, `opus[1m]` — return False and get routed through the runner instead of the direct Anthropic client.

Suggest broadening the alias match (prefix-match on `opus`/`sonnet`/`haiku`) so these classify as Anthropic.

agent_eval/events.py
```diff
@@ -159,6 +186,263 @@ def parse_stream_events(stdout_text, result_cap=DEFAULT_RESULT_CAP):
...
+        # Cursor's documented stream assistant message is intentionally lean.
+        return isinstance(message, dict) and not any(
+            key in message for key in ("id", "model", "usage"))
+    if event_type == "result":
+        # Cursor result events have duration/request metadata rather than the
+        # Claude usage/cost fields.  request_id is optional, duration_ms is not
+        # on the documented successful result shape.
+        return (isinstance(obj.get("result"), str)
+                and ("request_id" in obj or "duration_ms" in obj)
+                and "total_cost_usd" not in obj)
```
astefanutti, line 208: **Consistency (low).** Contradictory assumption about Cursor's `result` event shape: here `_looks_like_cursor_object` treats the presence of `total_cost_usd` as proof the event is **not** Cursor, but `cursor_agent.py::_parse_cursor_stream` reads `total_cost_usd` **from** the result event (`cost = obj.get("total_cost_usd")`). If Cursor's result event does carry cost, a stream distinguished only by that event fails detection and its terminal answer isn't surfaced. Worth reconciling which assumption is right against a real Cursor stream.

agent_eval/config.py
```diff
@@ -1629,14 +1629,17 @@ def from_yaml(cls, path: str | Path) -> "EvalConfig":
...
                          if runner.type == "codex"]
-        if codex_runners and config.inputs.tools:
-            raise ValueError(
-                "runner.type 'codex' does not support inputs.tools interception; "
-                "use claude-code or remove the tool interceptors")
-        if any(runner.workspace_mode == "repo" for runner in codex_runners):
-            raise ValueError(
-                "runner.type 'codex' does not support workspace_mode: repo "
-                "because repository answer-key protections cannot be enforced")
+        for runner in codex_runners:
```
astefanutti: **Nit (low, optional).** `inputs.tools` / `workspace_mode: repo` are rejected at config-load only for codex here; the cursor equivalent is enforced later in `CursorAgentRunner.from_config`, so a cursor config with `inputs.tools` loads clean and fails at runner-build time instead. Not blocking — a generalized per-runner capability check would cover both; fine to fold into the follow-up.

astefanutti (commented):
Thanks @Prashanth684, this is a great feature! I've left some review comments.

astefanutti:
Pushed two commits (`969a82d`, `96678f9`) addressing the review, plus a merge of `main` (`891c60d`) that resolves the conflict — the PR is mergeable again. All points from both reviews (my inline review + CodeRabbit) are addressed:

**Cursor runner**
- `allow` list mapping to zero Cursor patterns: kept the intentional fail-closed (deny-all) behavior but it now **warns** instead of silently locking the agent out.
- `per_model_usage`/`per_model_turns` now key by the same (assistant) model identity — no more split tokens vs turns.
- Cursor result events no longer read `total_cost_usd` (a numeric `total_cost_usd` marks a *Claude* result per the stream detector — the two were contradictory).
- Empty terminal Cursor `result` now falls back to the streamed deltas instead of surfacing an empty conversation.
- Reject `inputs.tools` for cursor at **config load** (same altitude as codex; `workspace_mode: repo` stays codex-only since cursor supports repo mode).

**Judge routing / security**
- `is_anthropic_model` prefix-matches aliases (`sonnet-4-5`, `opus[1m]`).
- `_load_llm_judge`: **keep the make_judge / OpenAI-compatible path** for non-Anthropic and credential-less Claude models — only a runner-managed model with *no external judge creds* falls through to the runner (fixes the regression that force-routed every non-Anthropic judge to the runner). Uses main's `_render_judge_prompt` and its correct `if jc.score_range` scale gate.
- `run_prompt_via_runner` strips `add_dirs` so a prompt-only judge grading untrusted output can't inherit host-dir grants (CWE-200/829), mirroring the agent-judge path.
- `_extract_agent_verdict` parses the whole/outermost JSON object before the legacy brace-free regex, so a valid verdict whose rationale contains braces isn't rejected.

**Docs**: incomplete Cursor permission-mapping disclosure (README/runner.md), local-only (not EvalHub/Harbor), `env` section + `--effort` levels made runner-specific, existing `.cursor/cli.json` caveat, cursor added to the intro runner list.

**Tests**: updated to the corrected behavior + added routing coverage; 1490 pass locally (the only failures are pre-existing env gaps — `pingouin` for anova, the `openai` package for the responses-api runner — unrelated to this PR).

**One deferral:** `max_budget_usd` is **not enforced by the Cursor CLI** (CodeRabbit's Major C12). I documented/commented the limitation rather than doing the heavy "reject Cursor judges / add external metering" — happy to add hard enforcement in a follow-up if you'd prefer it in this PR.

astefanutti:
Addressed the follow-up review:

- **`.eval-venv` committed (CWE-200)** — fixed in `e2ee9df`: untracked the symlink and hardened `.gitignore` (it only listed `.eval-venv/`, which doesn't match a symlink). My mistake via a `git add -A` in the merge commit; the net diff no longer contains it.
- **`_find_skill_text` symlink read (CWE-59, Major)** — fixed in `22e5812`: resolve the SKILL.md path and confine it to its skill root before reading, so a repo-planted `skills/<name>/SKILL.md` symlink can't leak a host file into the prompt. (Skill *names* were already `..`-guarded by `_skill_references`.) Added symlink-rejection + regular-file tests.
- **Docs** (`22e5812`) — scoped the `inputs.tools` handler-verification guidance to claude-code (codex/cursor reject it at load); noted in `CLAUDE.md` that cursor is local-only (no Harbor/EvalHub).

**Deferred (not this PR): `_FencedOutputs` structured-file fencing (CWE-1427, score.py:624).** `_FencedOutputs.__str__` fences only bare `{{ outputs }}`, not `outputs.files["…"]` or the safety-judge file loop. That's a gap in `main`'s untrusted-material-guard feature — not Cursor-runner code, and it surfaced here only because this PR merged `main`. It also affects every judge path independently of Cursor. Better tracked as a separate hardening against `main` (a render-time proxy for `outputs.files`) than bolted onto the Cursor PR. Happy to open that issue/PR separately.

astefanutti:
Ran a broader documentation audit against the PR's actual behavior and updated the docs (`3118964`) — 13 files across four themes:

- **Cursor is local-only** (base image ships no `cursor-agent`, so Harbor/EvalHub reject it): noted in `backends.md`, `deploy/README.md`, `guides/harbor.md`, `guides/evalhub.md`, `container-images.md` — several previously claimed the config "runs unchanged on Local/Harbor/EvalHub" with no exception.
- **Cursor env vars** (`CURSOR_API_KEY` / `CURSOR_API_ENDPOINT` / `CURSOR_AGENT_BIN`): added to `environment-variables.md` (they were absent) + generalized the env-allowlist note.
- **Judge backend routing** is three tiers (Anthropic SDK / MLflow `make_judge` / runner-backed), not "Anthropic or runner": corrected `reference/config/judges.md`, `concepts/judges.md`, `data-pipeline.md` (they omitted the `make_judge` tier and over-generalized non-Anthropic → runner).
- **Runner enumerations** that omitted `cursor`/`codex`: `backends.md`, `concepts/index.md`, `get-started/index.md`. Also added a dedicated `## cursor` section to `concepts/runners.md` and documented the fail-closed empty-allowlist behavior in `permissions.md`.

(Edits are text-only; the mkdocs strict build runs in CI.)

astefanutti (approved):
Thanks!

PR #214, opened 2026-09-01
fix(report): attribute the analysis to its author model, not the eval subject

astefanutti:
## Problem

The analysis callout's **"Generated by `<agent>` (`<model>`)"** line credits the **analysis author** (the eval-run orchestrator), rendered from the `analysis.md` YAML frontmatter. The template (`analyze-results.md`) said `model: <your-model-id>` ambiguously, so the agent recorded the **eval subject** (`--model` — the model under test) instead of *its own* model. The callout then reads as if the model under test authored the analysis. Real example (`rfe.speedrun`, run `2026-08-28-glm-5.3-flash-v2`, subject `z-ai/glm-5.3-flash`):

```
Generated by claude-code (z-ai/glm-5.3-flash)
```

— but the analysis was written by Claude; glm was only the subject (already shown in the report's **Model** row).

## Fix

The orchestrator agent already knows its own model (it's in its system prompt), so this is a wording bug, not a missing-data problem — and the fix must stay provider-neutral (the harness evaluates many models).

- **`analyze-results.md`** — clarify that `model` is the author's **own** model (the model *you* are running as, from your context, any provider), explicitly **not** `--model`/the subject; omit only if genuinely unknown. Includes the concrete failure it prevents.
- **`report.py`** — unchanged; it already renders the frontmatter as-is. (An earlier revision of this PR added run_result sourcing / a conflation guard; dropped in favor of trusting the now-unambiguous template, per review — it keeps the model visible in every case, including same-model runs, and stays provider-neutral.)
- **`tests/test_report_attribution.py`** — pin the render contract: frontmatter agent+model render verbatim, the subject model (`run_result.model`) never leaks into the author line, non-Anthropic author models render, and the agent-only / default fallbacks.

## Test

```
$ .venv/bin/python3 -m pytest tests/test_report_attribution.py tests/test_report_markdown.py tests/test_report_score_bands.py -q
22 passed
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved analysis attribution so reports identify the author’s agent and model accurately.
  - Prevented the evaluated subject model from being incorrectly displayed as the analysis author.
  - Added support for provider-neutral model identifiers and agent-only attribution when no model is available.
  - Added a sensible default attribution when author details cannot be determined.

- **Documentation**
  - Clarified how to specify the author’s own model identifier in analysis metadata.


astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

PR #215, opened 2026-09-01
fix(eval-analyze): derive eval.md name from config so flat layouts don't collide

astefanutti:
## Problem

Flat multi-eval (`eval/<name>.yaml`) is a first-class, tested layout — `discover_configs` derives a unique `eval_name` per config and even warns on name collisions (`test_flat_name_collision_warns`). But the `eval-analyze` cache was always written to `<config-dir>/eval.md` (fixed filename), so **multiple flat configs silently overwrite each other's `eval/eval.md`** — with no warning, unlike the config-name case.

Example: `eval/epic-decompose.yaml` and `eval/user-guides.yaml` both map their analysis cache to `eval/eval.md`.

## Fix

Derive the cache path from the config filename: **config path with `.yaml` → `.md`**.

- Root `eval.yaml` → `eval.md` (unchanged)
- Nested `eval/<name>/eval.yaml` → `eval/<name>/eval.md` (unchanged)
- Flat `eval/<name>.yaml` → `eval/<name>.md` (**fixed** — distinct per config)

Changes:
- **`agent_eval/config.py`**: add `analysis_cache_path(config_path)` returning the config path with a `.md` suffix, documenting the flat-collision rationale.
- **`skills/eval-analyze/SKILL.md`**: set `<eval_md_path>` to the config path with `.yaml` → `.md` (was: fixed filename `eval.md`), in both Config Location Discovery and Step 7.
- **`tests/test_discovery.py`**: cover root/nested (`eval.md`) and flat (distinct `<name>.md`, `a != b`).

## Impact

`eval.md` is a regenerable analysis cache read only by `/eval-analyze --update` (freshness) and `validate_eval.py` — **not** by `/eval-run`. So this doesn't affect running evals; it stops the analysis cache from clobbering across flat-layout evals.

## Test

```
$ .venv/bin/python3 -m pytest tests/test_discovery.py tests/test_layout.py tests/test_reorganization.py -q
32 passed
```

## Summary by CodeRabbit

- **New Features**
  - Eval analysis results now use uniquely named `.md` cache files derived from each configuration file.
  - This applies consistently to both discovered configurations and explicitly specified configurations.
- **Documentation**
  - Updated eval analysis guidance to reflect the revised output-path behavior.

### DIFF
--- a/agent_eval/config.py
+++ b/agent_eval/config.py
@@ -1834,6 +1834,21 @@ def discover_configs(project_root: Path) -> list[DiscoveryResult]:
     return sorted(results, key=lambda r: r.path)
 
 
+def analysis_cache_path(config_path) -> Path:
+    """Return the eval-analyze cache (``eval.md``) path for a config.
+
+    The cache lives next to the config, named by swapping the config's
+    extension to ``.md``: ``eval.yaml`` -> ``eval.md`` (root and nested
+    ``eval/<name>/eval.yaml`` layouts, unchanged), and flat
+    ``eval/<name>.yaml`` -> ``eval/<name>.md``. Deriving the name from the
+    config filename keeps flat-layout configs from silently colliding on a
+    single ``eval/eval.md``: flat configs are uniquely named (and warned on
+    name collision in ``discover_configs``), so their analysis caches must be
+    uniquely named too.
+    """
+    return Path(config_path).with_suffix(".md")
+
+
 def infer_layout(configs: list[DiscoveryResult]) -> str:
     """Infer the project's eval layout from discovery results.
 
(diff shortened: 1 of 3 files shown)


astefanutti:
@coderabbitai review

astefanutti:
Good catch, thanks @coderabbitai — fixed in a0cfccf. Step 7's generation instruction still hardcoded `Write eval.md`, which for a flat `eval/<name>.yaml` would have written the shared `eval/eval.md` and re-introduced the very collision this PR fixes. It now writes `<eval_md_path>`. Also updated the Step 8 report line (`eval.md` → `<eval_md_path>`) for the same reason. Verified `skillsaw --strict` passes and the file stays under the context-budget.

PR #216, opened 2026-09-02
feat(judges): cross-provider judges — decouple judge provider from runner (follow-up to #213)

astefanutti:
## Cross-provider judges — decouple the judge LLM provider from the eval runner

Follow-up to #213. Routes the **judge backend by the judge model's provider**, independent of `runner.type`, so an Anthropic judge works with a Cursor/Codex runner and an OpenAI (or OpenAI-compatible) judge works with a claude-code runner.

> [!IMPORTANT]
> **Stacked on #213 — the diff includes #213's commits as overlap.** This branch is based on #213's head (`agent-eval-harness-cursot`, from a fork), not `main`, because it extends the `prompt_backends.py` / runner-judge machinery #213 introduces. Review only the three follow-up commits:
> - `docs(spec): 013 cross-provider judges`
> - `feat(judges): route judge backend by model provider, independent of runner`
> - `fix(judges): strip provider prefix in all judge-model consumers; tighten validation`
>
> **Merge order:** land #213 first, then I'll rebase `--onto main` (the overlap drops out) for a clean single-purpose diff. Supersedes the minimal in-PR fix for #213 finding #1.

### What changes

The judge model carries the provider as a `provider:/model` id (or a bare id):

| Judge model | Backend |
| --- | --- |
| `sonnet`, `claude-sonnet-4-5`, `anthropic:/…` | Anthropic SDK |
| `openai:/gpt-4o`, `gpt-4o`, any other bare id | OpenAI SDK (`OPENAI_BASE_URL` reaches OpenAI-compatible gateways) |
| `runner:/gpt-5.4-medium` | configured runner (explicit opt-in for runner-only ids) |
| `gemini:/…` (explicit unsupported) | rejected at config load |

- `prompt_backends.py`: `split_model_uri`, URI/prefix-aware `is_anthropic_model` (fixes `sonnet-4-5`/`opus[1m]`), `resolve_judge_backend`.
- `score.py`: new `_call_structured_judge_openai` (function-calling, mirrors `_call_structured_judge` — rationale-first, bounds, images) and `_call_pairwise_openai`; provider dispatch across **all four** judge-model consumers (`_load_llm_judge`, `_make_builtin_scorer`, `_load_agent_judge`, pairwise). Removes the broken `make_judge` path.
- `config.py`: fail-fast load validation (skips agent judges); reject unsupported `feedback_type`.
- `generate_synthetic.py`: strip provider prefix; guard OpenAI-native generation.
[...]

PR #218, opened 2026-09-14
fix(judges): fence structured file access in judge prompts (CWE-1427)

astefanutti:
## Fence structured file access in judge prompts (CWE-1427)

LLM judge prompts fence untrusted, agent-produced content between
`[BEGIN/END EVALUATED MATERIAL]` markers that the judge system prompts are told
never to obey. But `_FencedOutputs` only fenced the **bare** `{{ outputs }}`
listing — templates that read a file directly rendered its content **unfenced**:

- `{{ outputs.files["result.md"] }}`
- the builtin safety judge's loop: `{% for path, content in (outputs.files or {}).items() %}{{ content }}`

So an evaluated agent could write instructions into a file and steer the LLM
judge's verdict (prompt injection, CWE-1427). Surfaced by CodeRabbit on #216 and
deferred there because it's a `main`-side hardening of the untrusted-guard
feature, independent of the Cursor runner / provider-routing work.

### Fix — a render-time proxy

- **`_FencedStr`** — a `str` subclass whose `str()` (what Jinja calls to emit
  `{{ x }}`) returns the fenced form, while comparisons, `in`, slicing, and
  `{% if %}` logic keep operating on the raw value.
- **`_FencedFiles`** — the `outputs.files` mapping: text values become
  `_FencedStr` on `[]` / `.get` / `.items` / `.values`; binary placeholders and
  non-string metadata pass through; iteration / `in` / `len` preserved.
- **`_FencedOutputs["files"]`** returns `_FencedFiles`.

This matches how `conversation` / `reasoning` / `inputs` / `tool_trace` /
`evidence` are already pre-fenced.

### Tests

Structured access and the file loop are now fenced; comparison/`in` logic still
sees the raw value; binary placeholders pass through. Full suite green (the only
failures are pre-existing env-gaps — `pingouin` for anova, the `openai` package
for the responses-api runner).

## Summary by CodeRabbit

* **Bug Fixes**
  * File contents accessed through structured outputs are now consistently wrapped with safety markers, including direct access and iteration.
  * File-content comparisons and membership checks continue to operate on the original values.
  * Binary file metadata remains available without text fencing.


astefanutti:
@coderabbitai review

astefanutti:
Addressed the filter-provenance finding (`46f44df`):

- **`| tojson`:** `_tojson_filter` now re-fences the serialized output when the value carries untrusted file content (`json.dumps` sees a `_FencedStr` as a plain str), so `{{ outputs.files[...] | tojson }}` and `{{ outputs.files | tojson }}` are fenced; trusted fields (`cost_usd`, …) are not over-fenced.
- **Moved fencing to Jinja's `finalize` hook** instead of `_FencedStr.__str__`. This fixes a subtler problem your finding surfaced: `| replace` previously operated on the *marker text* and corrupted it (`files` → `fil3s`). With `finalize`, string filters never see the markers — comparisons/`in`/`{% if %}` still use the raw value, and unfiltered content fences cleanly.

**Documented limitation (by design):** a string filter (`| upper`, `| replace`) returns a plain `str`, dropping the tag, so filtered file content is unfenced — value tainting can't follow arbitrary transforms without a full taint system (disproportionate here). Guarantee: **unfiltered file content is fenced**; templates shouldn't pipe untrusted file content through string filters. Tests cover tojson (fenced), trusted metadata (unfenced), and the string-filter limitation.

PR #219, opened 2026-09-15
fix(eval-analyze): detect stdout-only skills and use {{ conversation }} in judge prompts

### LINKED ISSUE #63: eval-analyze should recommend {{ conversation }} for stdout-only skills
## Problem

LLM judges using `{{ outputs }}` in their prompt template receive empty content for stdout-only skills (e.g. daily standup generator). This causes judges to score low or error because there's nothing to evaluate.

## Root Cause

This is a **config generation issue**, not a code bug. The template variables have clear semantics:

- `{{ outputs }}` — file artifacts (from `outputs[].path` directories)
- `{{ conversation }}` — the assistant's text responses extracted from the event stream

For stdout-only skills, the correct template is `{{ conversation }}`, but `/eval-analyze` generates judge prompts using `{{ outputs }}` regardless of skill type.

## Expected Behavior

`/eval-analyze` should detect when a skill is stdout-only (no `outputs[].path` configured, or the skill's primary output is conversational text) and use `{{ conversation }}` in the generated judge prompts instead of `{{ outputs }}`.

## Affected Code

- `skills/eval-analyze/` — skill analysis and eval.yaml generation
- `skills/eval-analyze/references/eval-yaml-template.md` — template reference (should document when to use each variable)

adam-d-young:
## Description
Fixes #63                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                  
`/eval-analyze` generates `{{ outputs }}` in LLM judge prompts regardless of skill type. For stdout-only skills (no file artifacts, e.g. daily standup generator), `{{ outputs }}` is always empty, causing judges to score every case as failure.

### Changes                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                              
- `prompts/analyze-skill.md`: Add `stdout_only:` field to the analysis YAML schema so the Explore agent explicitly classifies the skill's output modality                                                                                                     
- `SKILL.md` (Step 6): Add rule that when `stdout_only: true`, all LLM judge prompts must use `{{ conversation }}` and inline check judges must use `outputs.get("conversation", "")`                                                                         
                                                                                                                                                                                                                                                              
Both files are prompt-driven (no Python changes) — eval-analyze is an LLM-orchestrated skill, so the prompts are the implementation.                                                                                                                          
                                                                                                                                                                                                                                                              
### Note                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                              
This fix has a soft dependency on #94 / PR #98. Our change gets `{{ conversation }}` into the right judge prompts, but `{{ conversation }}` content itself is still polluted with raw tool calls until #98 lands. Both fixes are independently correct and can merge in either order.                                                        

## How Has This Been Tested?                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                    
Prompt-level change — verified the analysis schema and generation instructions are consistent with the existing template variable documentation in `eval-yaml-template.md` and `judge-prompt-template.md`. Full end-to-end testing requires running `/eval-analyze` against a stdout-only skill.                                                                       
                                                                                                                                                                                                                                                    
<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **Documentation**
  - Added guidance for evaluating skills that produce conversational output without file artifacts.
  - Updated analysis reports to identify whether a skill produces stdout-only output.
  - Clarified that stdout-only evaluations should use conversational output, while file-based evaluations continue using generated artifacts.
  - Expanded judge prompt guidance for selecting the appropriate output in stdout-only evaluations.
  - Improved wording and clarity throughout the evaluation instructions.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


skills/eval-analyze/SKILL.md
```diff
@@ -163,6 +163,7 @@ Combine the skill analysis (Step 4) and dataset exploration (Step 5) into a comp
 - **Schemas**: `dataset.schema` and `outputs[*].schema` drive the whole pipeline — be specific, use the real file/field names you observed. Mark inputs that reference external systems with `[EXTERNAL: System]` so `/eval-dataset` won't fabricate them.
 - **Permissions**: if the skill's `allowed-tools` includes `Skill`, add `"Skill"` to `permissions.allow` — nested skill calls fail silently in headless without it.
 - **Tool interception**: if the skill uses AskUserQuestion or calls external services (MCP tools/scripts), add `inputs.tools` entries (`match` = natural-language description, `prompt` = how to handle). AskUserQuestion answering uses `models.hook` + per-case `answers.yaml`.
+- **Stdout-only skills**: if the skill analysis has `stdout_only: true`, every LLM judge prompt MUST use `{{ conversation }}` as its artifact variable — not `{{ outputs }}`. For stdout-only skills `{{ outputs }}` is always empty, so any judge using it will receive no content and score every case as failure. This applies to all LLM judges; inline `check` judges should use `outputs.get("conversation", "")`.
```
astefanutti: Three tightenings on this bullet (the fix itself is sound):

**1. Scope to output/quality judges — not "all LLM judges".** "every LLM judge prompt MUST use `{{ conversation }}` … applies to all LLM judges" over-generalizes: a process/behavior judge (navigation, tool usage) should grade `{{ tool_trace }}` / `{{ evidence }}`. The absolute wording contradicts `analyze-skill.md`'s "include only the variable(s) the rubric actually grades" and its "`{{ tool_trace }}` for agent behavior" mapping — injecting `{{ conversation }}` into a tool-usage judge feeds it the wrong content.

**2. "`{{ outputs }}` is always empty" → "bare `{{ outputs }}`".** Only the bare variable renders empty (`_FencedOutputs.__str__` lists files only); `{{ outputs.conversation }}` and `{{ outputs | tojson }}` still carry the full record. As written, an author may think *all* `outputs` access is dead for stdout-only skills. (Same wording in the `stdout_only` definition in `analyze-skill.md` — tighten there too.)

**3. Mention builtin judges.** The rule covers LLM-prompt and inline `check` judges but not **builtin** judges — the top of the selection ladder in the adjacent "Judges" bullet. No live breakage (the shipped builtin LLM judges serialize conversation via `{{ outputs.conversation }}` / `{{ outputs | tojson }}`), but a future file-reading builtin selected for a stdout-only skill would reproduce #63 with no guidance to avoid it.

Suggested: scope the rule to output/quality judges, say "bare `{{ outputs }}`", and add that builtin judges must likewise read the conversation (e.g. `{{ outputs.conversation }}`) for stdout-only skills.

astefanutti (commented):
Reviewed #219. The core fix is **sound** — `{{ conversation }}` is a real LLM-judge template variable and inline `check` judges do receive a `conversation` key in `outputs`, so the #63 mechanism works. No correctness bugs; the inline comment has three doc-precision/scope tightenings worth folding in before merge.

**Follow-up (not this PR):** the fix is a per-skill documentation special-case — it relies on the analyzer classifying `stdout_only` correctly (an LLM judgment, easy to get wrong). `score.py` already has an analogous *harness-level* guard for `{{ reasoning }}` (it warns when a template references it but no event trace exists). A generalized detector — warn when a judge's chosen output variable renders empty across *all* cases — would catch the whole #63 class regardless of that classification, rather than depending on a prompt bullet. Worth a separate issue.

skills/eval-analyze/SKILL.md
adam-d-young: Appreciate the review -- see latest push for tighter language. I scoped the rule to judges that would normally use bare {{ outputs }}, called out builtins, and clarified "bare" in both files.

skills/eval-analyze/SKILL.md
```diff
@@ -163,6 +163,7 @@ Combine the skill analysis (Step 4) and dataset exploration (Step 5) into a comp
 - **Schemas**: `dataset.schema` and `outputs[*].schema` drive the whole pipeline — be specific, use the real file/field names you observed. Mark inputs that reference external systems with `[EXTERNAL: System]` so `/eval-dataset` won't fabricate them.
 - **Permissions**: if the skill's `allowed-tools` includes `Skill`, add `"Skill"` to `permissions.allow` — nested skill calls fail silently in headless without it.
 - **Tool interception**: if the skill uses AskUserQuestion or calls external services (MCP tools/scripts), add `inputs.tools` entries (`match` = natural-language description, `prompt` = how to handle). AskUserQuestion answering uses `models.hook` + per-case `answers.yaml`.
+- **Stdout-only skills**: if the skill analysis has `stdout_only: true`, any judge that would normally use bare `{{ outputs }}`, including builtins, MUST use `{{ conversation }}` instead (or `outputs.get("conversation", "")` for inline `check` judges) — bare `{{ outputs }}` renders empty for these skills. Judges using other variables (`{{ tool_trace }}`, `{{ evidence }}`, etc.) are unaffected.
```
astefanutti, line 166: **The failing `lint` check is skillsaw `--strict` (context-budget), not a logic problem** — `test` and CodeRabbit pass.

`skills/eval-analyze/SKILL.md` is **4,591 tokens, 91 over the 4,500 limit**; in `--strict` mode that single warning fails the job.

The catch: this file is **already at the ~4,500 ceiling on `main`** (it passes just under), so the new bullet has no room. Verified locally with skillsaw:
- tightening the new bullet alone → 4,569 (still 69 over)
- tightening the new bullet **and** the `Judges` bullet → 4,544 (still 44 over)

So ~90 tokens need reclaiming — more than the new bullet's own weight. Two realistic paths:

**1. Trim ~90 tokens across bullets.** A tighter new bullet (verified −22, keeps all the meaning):
> - **Stdout-only skills** (`stdout_only: true`): bare `{{ outputs }}` renders empty, so any judge that reads it — including builtins — must use `{{ conversation }}` instead (inline `check`: `outputs.get("conversation", "")`). Judges on other variables (`{{ tool_trace }}`, `{{ evidence }}`, …) are unaffected.

…plus tightening `Judges`/`Tool interception`/etc. to reach ~−90.

**2. (Recommended, sustainable) Relocate** the stdout-only guidance into `references/judge-prompt-template.md` — which the `Judges` bullet already points prompt authors to — leaving only a short pointer here, and shave a little existing content so the file nets under budget. Since the SKILL.md is at the ceiling, some content should move to a reference regardless.

Happy to push a complete, skillsaw-verified version if that's easier than iterating on the budget.

adam-d-young: I moved the guidance to `references/judge-prompt-template.md` as suggested.

I also trimmed several tokens mostly just trying to drop articles and compress parentheticals where possible without losing meaning. Feel free to push any further corrections/reductions.

astefanutti (approved):
Thanks @adam-d-young!

